PR #79, opened 2015-01-11
Changed to integer division to fix float error

xtream1101:
Fixed this error
`Traceback (most recent call last):
  File ".\archive.py", line 7, in <module>
    for result in search:
  File "C:\Programming\Python\Python34\lib\site-packages\internetarchive\search.py", line 70, in __iter__
    for page in range(1, total_pages):
TypeError: 'float' object cannot be interpreted as an integer`

Edited the way `total_pages` was calculated to use integer division


jjjake:
Thanks @xtream1101, I appreciate the contribution!

PR #95, opened 2015-09-30
Fixing tests

JesseWeinstein:
These tests fail (at least for me). These changes fix them. <s>There is still one failing test, `test_get_auth_config` which appears to be due to the test cookies not being accepted, which I haven't been able to figure out. :-( </s>

I have now worked around the bug(s) in the responses & requests packages such that ALL THE TESTS PASS.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
 pytest==2.6.0
 pytest-cov==1.8.1
-responses==0.3.0
+responses==0.4.0
 wheel
--- a/tests/cli/test_ia_search.py
+++ b/tests/cli/test_ia_search.py
@@ -1,4 +1,7 @@
-import json
+try:
+    import ujson as json
+except ImportError:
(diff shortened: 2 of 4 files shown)


JesseWeinstein:
I wanted to put them in the function so the Ugly Hack(TM) is locallized, and can be more easily taken out once the underlying issue is fixed.

JesseWeinstein:
It doesn't actually fix it. :-( Still figuring out what went wrong.

JesseWeinstein:
OK, now the test_config.py tests work -- but `test_ia_metadata_formats` fails, because @jjjake added two files to `iacli-test-item60` with `JSON` and `HTML` formats, respectively.

See https://archive.org/history/iacli-test-item60

jjjake:
@JesseWeinstein I've removed the JSON and HTML formats from `iacli-test-item60`.

PR #96, opened 2015-09-30
Add len support

JesseWeinstein:
I still need to add tests, but I don't want to do that until all the *existing* tests pass, which they don't (at least for me) (yet). But I thought I'd push this up, so you can at least see it.


JesseWeinstein:
OK, this has tests now. I think it's ready to be merged. Please let me know if you have any other concerns, and I'll do my best to resolve them.

PR #104, opened 2015-10-26
Add clue for folks running on thinly-configured systems

wumpus:


saper:
Well, Linux distributions are not a whole world....

wumpus:
Most Mac devs use ports, and the package names are similar. Windows, yeah, that's complicated.

saper:
I mean also this that even not all Linux distributions have a concept of "-dev" packages.

wumpus:
Uh, sure, there are an infinite number of Linux distros and some have different naming schemes. I think this short bit of advice works for the majority of devs... and at the least, it highlights the fact that there are dependencies to the OS which are not going to be automatically satisfied. Most python packages just work.

wumpus:
Whoops, meant to make this 2nd thing a separate pull request.

PR #112, opened 2015-11-22
Support search collection metadata

JesseWeinstein:


JesseWeinstein:
Replaced with PR #120

PR #116, opened 2016-01-06
Make Items hashable and orderable

JesseWeinstein:
Replacement for #113 , now rebased off current master branch.

Please don't close this one without at least some explanation. :confused:

### DIFF
--- a/internetarchive/item.py
+++ b/internetarchive/item.py
@@ -13,6 +13,8 @@ from fnmatch import fnmatch
 import os
 from time import sleep
 import sys
+import functools
+import json
 
 from six import string_types
 from requests import Response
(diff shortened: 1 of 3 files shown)


jjjake:
@JesseWeinstein I did not close your last pull request intentionally, my apologies!

Note that we are not testing against Python versions 2.6, 2.7, 3.4 and 3.5 (via tox), and also testing for pep8 compliance (with a few exceptions defined in https://github.com/jjjake/internetarchive/blob/master/setup.cfg). It looks like you have a fewer minor pep8 and Python 2.6 syntax issues to fix before the tests pass.

Can you fix these, so I can merge this PR? Thanks again! : )

PR #123, opened 2016-02-12
Add function to display recent items from uploaders with few other items

JesseWeinstein:


jjjake:
I'm closing this as I think something like this should be handled outside of the library. Please comment/reopen if you disagree Jesse.

JesseWeinstein:
It makes sense to put this outside the library, but I'm not sure where. Any suggestions?

jjjake:
@JesseWeinstein If you only need it in the CLI, perhaps it would make sense as a plugin?

See this repo as an example of how to do that: https://github.com/jjjake/ia_plugin

JesseWeinstein:
That does seem like a good way to do it. OK, I'll make such a thing. Should I put it in a separate repo, or is there a central repo that it would make sense for me to fork from?

jjjake:
@JesseWeinstein put it in a separate repo. Perhaps then we can add support for `pip install internetarchive[plugins]` to ia-wrapper.

JesseWeinstein:
OK, here it is: https://github.com/JesseWeinstein/ia_recent 

Lots more features to add, but it works.

PR #134, opened 2016-03-16
Read filename from file object with FS encoding

dsg22:
Fixes a crash when uploading files with non-ASCII filenames.


dsg22:
Nevermind, this is not ready for pulling as is. Sorry for the PR spam.

PR #150, opened 2016-08-09
Cast all headers to string in prepare_headers

craffel:
Some headers were being passed to requests with integer values, which would cause requests to raise a InvalidHeader error (at line [799 of requests/utils.py](https://github.com/kennethreitz/requests/blob/master/requests/utils.py#L799)).  This casts all headers to string, which solves the issue.  I tested this with an upload to the Internet Archive and it avoided the error for me - was not able to get uploading to work otherwise.


jjjake:
@craffel Thank you so much for the PR -- I actually just patched this before I saw your PR!

Thanks again though, 'preciate it! : )

craffel:
Hah nice!

Issue #151, opened 2016-08-10
Bug: delete --all fails on 1.0.8


vxbinaca:
```ia delete Snydsnap_Youtube_Archive --all```

```
Deleting files from Snydsnap_Youtube_Archive
 deleting: [Cover] Fly me to the moon-omVL1YATrwA.annotations.xml
Traceback (most recent call last):
  File "/usr/local/lib/python3.5/dist-packages/requests/utils.py", line 752, in check_header_validity
    if not pat.match(value):
TypeError: expected string or bytes-like object

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/bin/ia", line 11, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.5/dist-packages/internetarchive/cli/ia.py", line 155, in main
    sys.exit(ia_module.main(argv, session))
  File "/usr/local/lib/python3.5/dist-packages/internetarchive/cli/ia_delete.py", line 129, in main
    retries=args['--retries'])
  File "/usr/local/lib/python3.5/dist-packages/internetarchive/files.py", line 291, in delete
    prepared_request = request.prepare()
  File "/usr/local/lib/python3.5/dist-packages/internetarchive/iarequest.py", line 80, in prepare
    queue_derive=self.queue_derive,
  File "/usr/local/lib/python3.5/dist-packages/internetarchive/iarequest.py", line 96, in prepare
    self.prepare_headers(headers, metadata, queue_derive)
  File "/usr/local/lib/python3.5/dist-packages/internetarchive/iarequest.py", line 154, in prepare_headers
    super(S3PreparedRequest, self).prepare_headers(headers)
  File "/usr/local/lib/python3.5/dist-packages/requests/models.py", line 409, in prepare_headers
    check_header_validity(header)
  File "/usr/local/lib/python3.5/dist-packages/requests/utils.py", line 756, in check_header_validity
    "not %s" % (value, type(value)))
requests.exceptions.InvalidHeader: Header value 0 must be of type str or bytes, not <class 'int'>
```

saper:
Yes, this looks like a new check in 2.11.0

saper:
@vxbinaca can you see if #152 checks that one for you? (probably downgrading Python requests module would help, too).

vxbinaca:
What value do you want me to wrap in ```'```? The item name?

saper:
[...]

PR #152, opened 2016-08-11
Convert boolean value to the string explicitly

saper:
requests 2.11 is checking types of HTTP headers now (this could seen as a followup to https://github.com/jjjake/internetarchive/commit/76b5b16d1bd2c721f691ec7888d010cc27076b4a)

Fixes https://github.com/jjjake/internetarchive/issues/151

### DIFF
--- a/internetarchive/files.py
+++ b/internetarchive/files.py
@@ -260,7 +260,7 @@ class File(BaseFile):
                       without sending the delete request.
 
         """
-        cascade_delete = False if not cascade_delete else True
+        cascade_delete = '0' if not cascade_delete else '1'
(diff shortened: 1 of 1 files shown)


jjjake:
Thanks for catching and chasing this down @saper.

Would you mind changing https://github.com/jjjake/internetarchive/blob/master/internetarchive/files.py#L263 to something like: `cascade_delete = '0' if not cascade_delete else '1'`, rather than in the `S3Request()` call?

I'm happy to do that myself as well, just let me know.

`

saper:
Ouch, I should have noticed that. Fixed this and removed my ugly code.

jjjake:
Thanks @saper!

PR #165, opened 2016-11-20
Handle case where invalid media-type is returned.

brycedrennan:
`TypeError: unhashable type: 'list'` was happening if the metadata had a non-hashable mediatype.
See identifier `CIA-RDP96-00789R000700210007-5` for an example.

 - add test for this case. 
 - added myself to contributors section

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -8,6 +8,11 @@ Development Lead
 
 - Jake Johnson <jake@archive.org>
 
+Contributors
+------------
+
+- Bryce Drennan <internetarchive@brycedrennan.com>
+
 Patches and Suggestions
 -----------------------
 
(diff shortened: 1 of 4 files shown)


brycedrennan:
@jjjake I'm a little concerned that one of the tests failed and then passed. Has this happened before?

jjjake:
@brycedrennan I have had this exact failure happen to me recently (only on the py27 test as well).

I actually reran the job on Travis just now, and it passed successfully. I suggest we keep it in mind, but not worry about it, too, much right now. It seems to be transient.

As far as the changes go, please note that mediatype should always be hashable -- `['texts', 'texts']` is not a valid mediatype value. It should simply be `texts`. There are only 9 valid mediatypes: 

```audio
collection
data
etree
image
movies
software
texts
web```

See http://internetarchive.readthedocs.io/en/latest/metadata.html#mediatype for more details on mediatype.

I do think this change makes sense though, because you should still be able to initialize an Item object even if it has an invalid mediatype (e.g. for repairing the mediatype).

Thanks again for the contribution @brycedrennan, I really appreciate it!

brycedrennan:
Thanks!

PR #186, opened 2017-03-14
Fixes doc string on upload() method

mlissner:
Docs were OK, but docstring was not.


jjjake:
Thanks @mlissner, I really appreciate the contribution! : )

PR #201, opened 2017-07-10
Add the packaging metadata to build the ia snap

elopio:
This is a package for the secure installation of apps that works in most Linux distributions.
Landing it upstream will enable builds that then can be distributed to many users through the Ubuntu store.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,19 @@
+name: ia
+version: master
+summary: A Command-Line Interface to Archive.org
+description: |
+  This package installs a command-line tool named ia for using Archive.org
+  from the command-line.
+
+grade: devel # must be 'stable' to release into candidate/stable channels
+confinement: strict
+
+apps:
+  ia:
+    command: ia
+    plugs: [network, home, removable-media]
+
+parts:
+  internetarchive:
+    source: .
+    plugin: python


elopio:
Hello!

Thanks a lot for your project. I've suffered uploading stuff to the internet archive since they removed ftp. With your tool things are looking a better now. *hugs*

I work at Ubuntu, in the snapcraft project. This is a new packaging and delivery system to make it a lot easier and safer to put software in the Ubuntu store. You can read more about it here: http://snapcraft.io/

To give this a try in an Ubuntu 16.04 machine:

    $ sudo apt install git snapcraft
    $ git clone https://github.com/elopio/internetarchive
    $ cd internetarchive
    $ git checkout snapcraft
    $ snapcraft
    $ sudo snap install *.snap --dangerous

If you have any question, please let me know.

jjjake:
It's great to hear the ia tool has been helpful to you, @elopio!

Thanks for the PR as well. This is the first I've heard snapcraft, but it looks interesting. It looks like this would install both the `ia` command-line tool as well as the Python library? If both, I would suggest we name the package `internetarchive` rather than `ia`. If it's just installing the command-line program, `ia` makes sense.

It also looks like the name `ia` is already reserved (was that you?).

I'm happy to add support for building snaps, it seems easy enough. I can automate the building and publishing of the snap in my release workflow. How does that sound, or would you like to maintain the `internetarchive` snap package? That's fine with me as well, but it might be more complicated coordinating releases.

Thanks again @elopio, I appreciate it! : )

elopio:
Hey @jjjake, thanks for reviewing. 

The python modules are inside the .snap, but these packages are more intended for final users than for developers. A developer would install your library from pip in a virtualenv and use the python interpreter. That's doable from the snap, but with multiple limitations because for security the package is not installed like traditional debs are. It is just mounted read-only in the `/snap/` dir. That makes it a little complex to write a script that combines your modules and other third party modules. It's doable, but definitely a virtualenv is better for that. The snap is for the command line tool.

I didn't reserve the name. As a safe guard to avoid people taking ownership of existing projects in the ubuntu archive, all the names of deb binaries are reserved. Apparently, [duckduckgo has a binary called ia](https://packages.debian.org/stretch/amd64/python-duckduckgo2/filelist). That's not a big deal, if you want ia you just have to fill the registration form explaining your claim to the name, and we'll assign it to you. Or we can rename the snap to internetarchive as you mentioned, and ia can be an alias ([work in progress](https://forum.snapcraft.io/t/improving-the-aliases-implementation/18)). This second alternative doesn't require any approvals from store admins.

One of the ideas of snaps is to avoid the overhead of distro maintainers. So it would be great if you own the snap in the store. You can automate the continuous delivery with a couple of clicks in https://build.snapcraft.io or you can use [travis](https://snapcraft.io/docs/build-snaps/ci-integration) or something else. I'd be happy to help you there if you need a hand.

And the project is in active development. We are very eager to hear from people using snaps, so if there's a feature you are missing or something that could be easier, please let us know and we'll put it in our backlog.

pura vida

jjjake:
Got it. That all makes sense to me, and I think I'm understanding the purposes of a snap better. One of the big goals of the `ia` tool is to enable folks who aren't very technical to upload, download, and modify metadata in bulk. So, I think snap sounds like a good solution for "final users" on Linux. Let's not worry about the Python library in this context.

I put a request in for the "ia" name. It'd be great if we could get that for this snap, but `internetarchive` will work if not.

I'll probably just automate the snap building and publishing in my Makefile, so I can have tighter control over the releases (e.g. I don't want a new snap build/release every time I push to master.).

I'll merge this now.

elopio:
Great, thanks for the merge. I checked the store, and somebody already approved your name.

It's of course ok if you prefer to keep the delivery in your own scripts. Just note that the store has four channels: edge, beta, candidate and stable. build.snapcraft.io only pushes to the edge channel, that is supposed to be used by early adopters and contributors wanting to help testing. The snap doesn't appear on user searches until you move it to the stable channel. I mention this in case you were worried about users getting an unstable revision without noticing. But as everything with snaps, whatever makes you happy works for us :)

PR #206, opened 2017-07-24
Fixes #205 by adding request_kwargs param

### LINKED ISSUE #205: No timeout for s3_is_overloaded
I'm currently uploading about 3.4M items to the Internet Archive as part of the [RECAP Project][recap]. 

As I'm uploading, I've noticed my program freezing without recovery, which in the past has been because somebody (usually me) didn't remember to set the timeout on Python requests. [This drives me nuts][r], and in this instance, it appears to be because `s3_is_overloaded` [doesn't set a timeout](https://github.com/jjjake/internetarchive/blob/e9a319f7e6334ff766f1a9da5d18742b6f106e74/internetarchive/session.py#L331). `s3_is_overloaded` also doesn't allow the user to set a timeout manually, so there's no easy fix on my end. I'm on version 1.6.0, but it looks like this is a problem on master as well. 

Two thoughts:

1. Is it possible to get a 1.7.0 release with this fix? 

2. If 1.7.0 isn't ready, is it possible to get a 1.6.1 release with this fix? 

When these timeouts are missing, it freezes any program that relies on the code, possibly forever. It's a big issue. This could be resolved by adding a default timeout (as is elsewhere in the code), and I don't think it would cause any changes to the API that wouldn't be backwards compatible. 

Thank you very much.

[recap]: https://free.law/recap/
[r]: https://github.com/requests/requests/issues/3070

mlissner:
Pretty simple tune up to add `request_kwargs` as a param to the `s3_is_overloaded` method. Follows the pattern set by the other methods, and sets a default timeout of 12 seconds, as in the `get_metadata` method.

### DIFF
--- a/internetarchive/session.py
+++ b/internetarchive/session.py
@@ -321,14 +321,17 @@ class ArchiveSession(requests.sessions.Session):
         else:
             return _catalog.tasks
 
-    def s3_is_overloaded(self, identifier=None, access_key=None):
+    def s3_is_overloaded(self, identifier=None, access_key=None, request_kwargs=None):
         u = '{protocol}//s3.us.archive.org'.format(protocol=self.protocol)
         p = dict(
             check_limit=1,
             accesskey=access_key,
             bucket=identifier,
         )
-        r = self.get(u, params=p)
+        request_kwargs = {} if not request_kwargs else request_kwargs
+        if 'timeout' not in request_kwargs:
+            request_kwargs['timeout'] = 12
+        r = self.get(u, params=p, **request_kwargs)
         try:
             j = r.json()
         except ValueError:


jjjake:
Ah, I didn't see this before I pushed my fix! Thanks for the PR,  @mlissner!

I'm happy to surface `request_kwargs` in the method args, but forgot to do that myself. I don't think it will be very useful, but am happy to do that if you'd like.

mlissner:
I debated it. Seems better to be consistent...let the caller override if they want to. I saw at least one old PR/issue where somebody wanted to tweak this on some other method, I think. I'm happy either way though, personally.

jjjake:
I went ahead and added it: 85d1878ac4ad54d02a0ad093b90d4bf9426b4c4c

Thanks again for the contributions (both code & uploads)! : )

PR #209, opened 2017-08-07
Add the install instructions for the snap

elopio:


jjjake:
Thanks again, @elopio!

elopio:
Thanks to you @jjjake!

I'm really loving this, specially with a loop and --delete. If you add resumable uploads, I'll owe you way too many beers :)

We have a very popular blog at https://insights.ubuntu.com/. I think it would be great to have a post there talking about internet archive, and the ia snap. What do you think? Would you like to write it?

pura vida.

PR #258, opened 2018-06-26
Add --no-change-timestamp flag

JesseTG:
Right now, when a file is downloaded its timestamp on the file system is changed to reflect that on the Internet Archive.  This optional flag prevents that.

I added this because I'm downloading files onto an HPC cluster's scratch space.  Every day, the cluster deletes all files that haven't been modified in 30 days.  That implies that anything I download from the Internet Archive will be deleted unless I babysit the download and change the time stamp at the end.


jjjake:
Looks great, thanks for the contribution, @JesseTG!

JesseTG:
Happy to help, thanks for the tool.

PR #272, opened 2018-10-04
mention collections

anarcat:
The collection is another item that cannot be changed after upload. Mention it in a more obvious warning in the upload docs.


jjjake:
Thanks @anarcat!

PR #274, opened 2018-10-17
add progress bar to file hashing, speed up file hashing

ahundt:
We added a couple of minor fixes so we can see hashing progress & increased the chunk size to speed up hashing substantially.

### DIFF
--- a/internetarchive/item.py
+++ b/internetarchive/item.py
@@ -51,6 +51,18 @@ from internetarchive.files import File
 from internetarchive.iarequest import MetadataRequest, S3Request
 from internetarchive.utils import get_s3_xml_text, get_file_size, is_dir
 
+# Progress bars using https://github.com/tqdm/tqdm
+# Import tqdm without enforcing it as a dependency
+try:
+    from tqdm import tqdm
+except ImportError:
+    print("tqdm is not available. Progress bar functionalities will be disabled.")
+
+    def tqdm(*args, **kwargs):
+        if args:
+            return args[0]
+        return kwargs.get('iterable', None)
+
 log = getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


internetarchive/item.py
```diff
@@ -51,6 +51,16 @@
...
+# Progress bars using https://github.com/tqdm/tqdm
+# Import tqdm without enforcing it as a dependency
+try:
+    from tqdm import tqdm
+except ImportError:
+    def tqdm(*args, **kwargs):
+        if args:
+            return args[0]
+        return kwargs.get('iterable', None)
+
```
anarcat: it might be worth doing that magic only once, in one location, otherwise bitrot might break it.... maybe we could import from utils here instead?

internetarchive/utils.py
```diff
@@ -86,7 +96,7 @@ def norm_filepath(fp):
 def get_md5(file_object):
     m = hashlib.md5()
     while True:
-        data = file_object.read(8192)
+        data = file_object.read(65535)
```
anarcat: why 64k instead of 8k? it would seem important to document the reason for the change at least  least in the commitlog or better, in the source code.

internetarchive/utils.py
```diff
@@ -234,15 +244,19 @@ def recursive_file_count(files, item=None, checksum=False):
             except (AttributeError, TypeError):
                 is_dir = False
         if is_dir:
-            for x, _ in iter_directory(f):
-                lmd5 = get_md5(open(x, 'rb'))
+            pb = tqdm(iter_directory(f), ascii=True,
+                      desc='Checking md5 for duplicate files')
+            for x, _ in pb:
+                with open(x, 'rb') as f_handle:
+                    lmd5 = get_md5(f_handle)
```
anarcat: is it really worth doing a progress bar on md5 checks? how often is that slow enough to warrant such a change?

and doesn't it overlap with the previous progress bar?

internetarchive/item.py
```diff
@@ -815,7 +825,9 @@ def upload(self, files,
                 if isinstance(f, tuple):
                     remote_dir_name = f[0].strip('/')
                     f = f[-1]
-                for filepath, key in iter_directory(f):
+
+                pb = tqdm(iter_directory(f), ascii=True, desc='Uploading file')
+                for filepath, key in pb:
```
anarcat: what does this progress bar actually do?

internetarchive/utils.py
Walter-o: Reading in larger chunks is faster.
There is a point of diminishing returns and i don't know that point but it's at least above 1/3 of hard-drive speed

PR #277, opened 2018-10-22
Close files after opening.

Arkiver2:


kpcyrd:
@jjjake One of my uploads is incomplete due to this bug. The fix is a really minor change, any updates?

jjjake:
@kpcyrd Thanks for the poke. I just fixed it here (rather than dealing with the conflicts): https://github.com/jjjake/internetarchive/commit/4e4120f07c98ea98c61791293835df2797bfee61.

It'll be in the next release. How is this preventing your upload though, out of curiosity?

kpcyrd:
Awesome, thanks! `ia` runs out of file descriptors after a while. My last upload was a folder with a large number of files and `ia` crashed half-way through, `ulimit -n` is set to 1024 and the upload on archive.org now has exactly 1023 files in it.

Issue #293, opened 2019-03-23
upload command's status check flag doesn't check if item is full


AGSPhoenix:
```
phoenix@debian:~$ ia upload <FullItem> --status-check
success: <FullItem> is accepting requests.
phoenix@debian:~$ dd if=/dev/zero bs=1M count=1 of=testfile
1+0 records in
1+0 records out
1048576 bytes (1.0 MB, 1.0 MiB) copied, 0.00139101 s, 754 MB/s
phoenix@debian:~$ ia upload <FullItem> testfile
<FullItem>:
 uploading testfile: [################################] 1/2 - 00:00:00
 error uploading testfile: Access Denied - This item total number of bytes(1134253428522) is over the per item size limit of 1099511627776. Please contact info@archive.org for help fitting your data into the archive.
1 phoenix@debian:~$
```
Running version 1.8.1.

Can the status check be made to check the item size limit as well?

Issue #290, opened 2019-01-24
Broken pipe error when item is over size limit


Arkiver2:
When an item, in this case https://archive.org/details/files.pushshift.io_201812, is over it's size limit, the following error is returned:
```
requests.exceptions.ConnectionError: ('Connection aborted.', BrokenPipeError(32, 'Broken pipe'))
```

However, the error returned by Internet Archive is
```
<Error><Code>AccessDenied</Code><Message>Access Denied</Message><Resource>This item total number of bytes(1116485106904) is over the per item size limit of 1099511627776. Please contact info@archive.org for help fitting your data into the archive.</Resource><RequestId>920e7af2-677d-4f9b-ad9d-1681b53db715</RequestId></Error>
```
which is not shown to the user who got the broken pipe error.

Related is the following issue for `requests` https://github.com/requests/requests/issues/2422 for broken pipe errors.

AGSPhoenix:
I got bitten by this. Been trying to upload ~2 TB of old archival data, and I've been stuck waiting for the past few days hoping it was just an intermittent issue that would go away after a while. For large files, the upload will run for 30 seconds before failing with the broken pipe message. I saw the same timings regardless of connection speed.

I finally thought to upload a small test file just to see what would happen, and got the error message once the upload finished. Frustratingly, the `--status-check` flag doesn't indicate any problems with the item.

AGSPhoenix:
@jjjake, I noticed you've been making a few commits recently. Are you planning on working on some of the open issues? If not, would you accept PRs for them?

jjjake:
@AGSPhoenix PR are always welcome.

PR #297, opened 2019-04-21
Improve handling of full items

AGSPhoenix:
Show a warning message when an upload fails due to a ConnectionError, as is the case when the item is full. `--status-check` now checks for this. These two changes make the underlying issue in #290 more clear.

This is just about the minimum useful PR I could put together in a short time. I'm not really familiar with all the different uses of `ia`, so wasn't really sure how to properly signal the failure of a single upload for bulk uploads and the like. I'd like to improve on this later if time allows, but this is better than nothing.

### DIFF
--- a/internetarchive/cli/ia_upload.py
+++ b/internetarchive/cli/ia_upload.py
@@ -61,7 +61,7 @@ from copy import deepcopy
 
 import six
 from docopt import docopt, printable_usage
-from requests.exceptions import HTTPError
+from requests.exceptions import HTTPError, ConnectionError
 from schema import Schema, Use, Or, And, SchemaError
 
 from internetarchive.cli.argparser import get_args_dict, convert_str_list_to_unicode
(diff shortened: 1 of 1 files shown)


AGSPhoenix:
Oh yeah, fixes #293 as well

internetarchive/cli/ia_upload.py
```diff
@@ -176,6 +181,10 @@ def main(argv, session):
                   'Expect 503 SlowDown errors.'.format(args['<identifier>']),
                   file=sys.stderr)
             sys.exit(1)
+        elif get_item('{0}'.format(args['<identifier>'])).item_size >= 1099511627776:
```
anarcat: could we use a global, in capitals, at the top of the file here instead of this magic number? something like:

```
IA_MAX_SIZE = 1099511627776
```

or even better, what the actual value *means*:

```
# one tibibyte
IA_MAX_SIZE = 2^40
```

jjjake: I agree with @anarcat. Also, we already initialize the item on https://github.com/jjjake/internetarchive/blob/master/internetarchive/cli/ia_upload.py#L183, this check should happen after that so we only have to do it once.

Issue #315, opened 2019-11-24
Stop using MD5?


commutateigh:
It looks like IA items provide CRC32, MD5, and SHA-1 hashes, which are all insecure hash functions. I suggest that we use the least insecure (SHA-1) instead of MD5 for checking integrity. If it's there, you might as well use the best one, right?

Since all those functions aren't great, another option might be to use all of them at once, and maybe the size of the file as well. That would probably be the most robust unless the IA starts using non-archaic hash functions, [which is unlikely](https://help.archive.org/hc/en-us/articles/360016553532-Live-Music-Archive-etree-org-#checksum-file): 

> Systems have improved to the point where we no longer see checksums as necessary to verify accurate upload/download.

TL;DR: Abandon MD5, adopt SHA-1.

JustAnotherArchivist:
While collision resistance of MD5 and SHA-1 is broken, it doesn't matter here. The hashes are only to check against corruption during transfer or storage. Random corruption due to equipment or software errors are extremely unlikely to not alter the hash, to the point that it's not even worth considering. Malicious modification while keeping the hash fixed is infeasible as well; nobody has broken the hash functions' preimage resistance yet. Even MD4 and MD2 aren't considered broken in that respect.

You're many orders of magnitude more likely to calculate a hash for corrupt data due to a hard drive or memory bit flip than to have a matching hash but the wrong data.

jjjake:
I agree with what @JustAnotherArchivist said, and this would need to be changed at the IA-S3 API level. Thanks for the feedback though, @commutateigh!

PR #316, opened 2019-11-24
Abandon MD5, adopt SHA-1

commutateigh:
Tentatively implements #315 (use SHA-1 instead of MD5 everywhere where possible).

Turns out it's not as easy to switch to SHA-1 as I thought, mainly because of the very annoying `Content-MD5` HTTP header. Right now there are two separate functions `get_sha1` and `get_md5`. This means that when uploading files, if `verify` is set or the local files are deleted after upload, then each file will be read twice, once for SHA-1 and once for MD5. The hashes could be calculated at the same time. Please advise if this is a good idea.

### DIFF
--- a/docs/source/quickstart.rst
+++ b/docs/source/quickstart.rst
@@ -151,7 +151,7 @@ By default, the :func:`download <internetarchive.download>` function sets the ``
      skipping nasa/nasa_archive.torrent, file already exists based on length and date.
      skipping nasa/nasa_files.xml, file already exists based on length and date.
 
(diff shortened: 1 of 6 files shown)


saper:
Unless there is a server-side support for hashes other than MD5 (I don't know really) this is not really possible?

jjjake:
@saper is right, this is a server-side feature. That point, along with @JustAnotherArchivist's comments on #315, I'm going to close this.

I appreciate the contribution though, @commutateigh, thank you!

Issue #295, opened 2019-03-31
Mistake in 'requires' statement: backports.csv is only defined for python-2x


yurivict:
```
    install_requires=[
        'requests>=2.9.1,<3.0.0',
        'jsonpatch>=0.4',
        'docopt>=0.6.0,<0.7.0',
        'clint>=0.4.0,<0.6.0',
        'six>=1.0.0,<2.0.0',
        'schema>=0.4.0',
        'backports.csv < 1.07;python_version<"2.7"',
        'backports.csv < 1.07;python_version<"3.4"',
        'backports.csv;python_version>="2.7"',
        'backports.csv;python_version>="3.4"',
        'total-ordering;python_version<"2.7"',
    ],  
```

PR #319, opened 2019-12-09
remove backports requirement for newer python releases

anarcat:
The backports.csv requirement makes no sense in Python releases after
3.4: it's part of the standard library and works fine. Remove that
requirement from 3.5 and later, as it breaks the install on Debian and
probably other platforms.

Closes: #295


jjjake:
Thanks @anarcat!

PR #322, opened 2019-12-29
fix function name typo

miku:


jjjake:
Thanks for catching that, @miku!

PR #324, opened 2020-01-19
Read from ~/.config/ia.ini with fallback to ~/.ia regardless of the existence of ~/.config

### LINKED ISSUE #323: Fails to read configuration if ~/.config directory is created after 'ia configure'
If you run `ia configure` without `~/.config` existing, the configuration will be written to `~/.ia` instead. However, the way `ia` decides where to read from is checking for the existence of `~/.config` rather than the config file itself. As a result, if `~/.config` is created *after* running `ia configure`, the config reading and therefore nearly all `ia` commands fail until you either manually move the config file or rerun `ia configure` (resulting in two config files).

JustAnotherArchivist:
Fixes #323

### DIFF
--- a/internetarchive/config.py
+++ b/internetarchive/config.py
@@ -103,11 +103,9 @@ def parse_config_file(config_file=None):
     config = configparser.RawConfigParser()
 
     if not config_file:
-        config_dir = os.path.expanduser('~/.config')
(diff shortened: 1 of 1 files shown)


jjjake:
Thank you, @JustAnotherArchivist!

PR #335, opened 2020-04-10
puts up a progress bar during file downloads

mecolosimo:
Here is a small modification that uses tqdm's pbar to display a progress bar during downloads when given the `-v` option. Otherwise, it looks just like normal.

### DIFF
--- a/internetarchive/files.py
+++ b/internetarchive/files.py
@@ -31,6 +31,7 @@ import sys
 import logging
 import socket
 
+from tqdm import tqdm
 import six.moves.urllib as urllib
 import six
 from requests.exceptions import HTTPError, RetryError, ConnectTimeout, \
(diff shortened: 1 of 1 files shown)


internetarchive/files.py
```diff
@@ -257,6 +258,11 @@ def download(self, file_path=None, verbose=None, silent=None, ignore_existing=No
                 and return_responses is not True:
             os.makedirs(parent_dir)
 
+        pbar = tqdm(total=self.size, 
+                    unit='B', 
+                    unit_scale=True,
```
cclauss, line 263:
```suggestion
        pbar = tqdm(total=self.size,
                    unit='B',
                    unit_scale=True,
```
Fixes https://travis-ci.org/github/jjjake/internetarchive/jobs/673581805#L281

JustAnotherArchivist:
This has been rendered obsolete by the recently merged #488.

cclauss:
Maybe should have been rich because lots could be done with nicer UI.
https://www.willmcgugan.com/blog/tech/post/progress-bars-with-rich/

PR #348, opened 2020-05-26
replaced sys.*.write calls with print

Dobatymo:
- consistency
- this also fixes some tiny output formatting errors (like mixed str/bytes)

For example in ia_delete this resulted in output like "asd/b'qwe'"

### DIFF
--- a/internetarchive/cli/ia_delete.py
+++ b/internetarchive/cli/ia_delete.py
@@ -96,7 +96,7 @@ def main(argv, session):
         args['--header']['x-archive-keep-old-version'] = '1'
 
     if verbose:
-        sys.stdout.write('Deleting files from {0}\n'.format(item.identifier))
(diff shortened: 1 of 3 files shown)


jjjake:
Thanks again for the contribution @Dobatymo!

This sounds good to me, but can you add `from __future__ import print_function` at the top of the scripts? Without it, these changes won't work in Python 2. For example:

```» python2.7 -c 'print("hi", file=sys.stderr)'
  File "<string>", line 1
    print("hi", file=sys.stderr)
                    ^
SyntaxError: invalid syntax
» python2.7 -c 'from __future__ import print_function; import sys; print("hi", file=sys.stderr)'
hi```

Dobatymo:
The three files I modified already have the `from __future__ import print_function` on top. Or am I missing something?

jjjake:
Ah, my mistake! Thanks again.

PR #349, opened 2020-05-27
add explicit --no-backup flag to ia delete

Dobatymo:
I think it's nicer to have an explicit flag to turn off backups.

Also I think the only two useful headers for _delete_ are `x-archive-keep-old-version` and `x-archive-cascade-delete`. Both are now available by flag.

Maybe the header options should be removed altogether and the functionality moved to `File.delete()`, which already has a `cascade_delete` argument.


jjjake:
This is great, thanks @Dobatymo!

I'll try to get this change added to `ia move`, `ia copy` and `ia upload` as well soon.

Issue #288, opened 2019-01-22
Derive queued before the last file (due to miscounting skipped files?)


JustAnotherArchivist:
I'm currently uploading a large dataset (159 WARCs, just over 1 TiB) using `ia upload`. Because of the general stability of IA's S3 interface (#176 et al.), I was expecting from the beginning that I'd have to run `ia` multiple times. However, my expectation was also that the derive would only be queued after the last file was uploaded. This, however, did not happen: while the second (first resumed) upload was still running, a derive task suddenly started.

A first `upload` process died with an `socket.timeout: The write operation timed out`/`Connection aborted` error after about three days and a bit over a third of the data (62 files) uploaded. I restarted it with `--checksum` to skip the files that were uploaded already. After checksumming everything again, it skipped the existing files and resumed the uploads starting with the 63rd file during which the first process had crashed. All fine so far, only archive.php and book_op.php tasks.

Yesterday (21st) around 13:00 UTC, while that second process was still running, a derive task (ID 1111473727) was suddenly queued and started. I don't know exactly how much data was uploaded at the time, but it looks like it was uploading the 96th file. Definitely still far from done. The derive was queued from an archive.php task which had `next_cmd=derive` set, unlike all the previous archive.php tasks. Since then, every completed upload restarted the derive task...

I can't help but notice that 62 + 96 is very close to the total of 159 files. Is the condition for setting the queue_derive flag perhaps not taking into account skipped files?

https://github.com/jjjake/internetarchive/blob/d093cdeeacd9f5a3beacd6fbacf6b2036560bc70/internetarchive/item.py#L820-L825

Issue #252, opened 2018-05-12
Feature request: support `queue_derive` in `modify_metadata`


p8NG6L:
Here's my use case: I have an item to upload with multiple files and metadata that varies per file. So I'm calling `upload(id, files=files, metadata=metadata)` once with the item metadata, then repeatedly calling `modify_metadata(id, target="files/<filename>", metadata=<file_metadata>)` for each of the files. I'd like to not start a derive until the entire process is finished. I can set `queue_derive=False` on the `upload` call, but `modify_metadata` doesn't take a `queue_derive` or `headers` argument, so there's no obvious way to start the derive task short of re-uploading one of the files (e.g., the smallest one).

I would also be happy if someone can suggest some no-op operation that would allow me to queue a derive as a side effect.

p8NG6L:
`Item.derive` in [v1.9.3](https://github.com/jjjake/internetarchive/releases/tag/v1.9.3) and the `ia <identifier> --cmd derive.php` command mentioned at https://github.com/jjjake/internetarchive/pull/351#issuecomment-637004784 solve this issue for me.

PR #351, opened 2020-05-28
WIP fix derive logic and redundant hash calc

Dobatymo:
This addresses https://github.com/jjjake/internetarchive/issues/253 and https://github.com/jjjake/internetarchive/issues/288

Counts all files (even those which might be skipped). This is consistent with the newly added behavior for file-level metadata.

BUG: if the last file is skipped, then a derive is never queued

Any idea how to solve this? Is there a way to queue a derive WITHOUT uploading a file?
The easiest way would be to just upload everything and skip whatever you like, and then after everything is completed, simple do queue a derive with a separate call.
Then we could do away with this fragile file counting.


JustAnotherArchivist:
> Is there a way to queue a derive WITHOUT uploading a file?

Related: #252

This would be nice to have in general. I wrote a script for myself last year since I didn't want to click around on the web interface all the time as I'm using `--no-derive` to avoid hitting those two bugs you're trying to solve as well as wanting to be sure everything's fine before initiating the derive that can take a very long time. I'm not sure if that works for non-admin accounts though, and it should most definitely not be used as a basis for an implementation here as it emulates the website interaction, so I'm not going to link it here.

jjjake:
Yes, it's possible to queue a derive task without uploading anything:

```$ ia tasks jj-test-2020-05-14 --cmd derive.php```

Python:

```>>> r = item.derive()```

Dobatymo:
Great, I will use this instead of the counting logic.

Dobatymo:
I removed all counting logic and just call derive() in the end. Do you think that is an acceptable solution?
The tests don't seem to like it though. If you think we should chose this solution. I will have a look at the tests.

JustAnotherArchivist:
@jjjake 

> --cmd derive.php

Could this be added to the documentation? It only mentions `make_dark.php` and `make_undark.php` currently: <https://archive.org/services/docs/api/tasks.html#supported-tasks>. Maybe also an example in the `ia tasks` help.

jjjake:
It's in the `ia tasks` help menu:

```...
examples:
    ia tasks nasa
    ia tasks nasa -p cmd:derive.php  # only return derive.php tasks
...```

I'll see about getting it added to the Tasks API documentation (I'm not in charge of that). Thanks for the heads up.

soredake:
What's the status of this?

Dobatymo:
@soredake jjjake didn't like the approach to have a seperate API call to queue the derive. Also at the time the unittests failed for some reason. I haven't looked at this PR for the past few years and now there are merge conflicts.

PR #366, opened 2020-08-04
Travis CI: Start testing on Python 3.9 beta in allow_failures mode

cclauss:
Warning from pytest running against OpenLibrary
```
/home/openlibrary/.pyenv/versions/3.8.5/lib/python3.8/site-packages/internetarchive/utils.py:33
  /home/openlibrary/.pyenv/versions/3.8.5/lib/python3.8/site-packages/internetarchive/utils.py:33: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will stop working
    from collections import Mapping
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,4 +6,8 @@ python:
     - "3.6"
     - "3.7"
     - "3.8"
+    - "3.9"
+jobs:
+  allow_failures:
+    - python: "3.9"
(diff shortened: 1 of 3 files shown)


jjjake:
Thanks @cclauss!

PR #371, opened 2020-09-08
Travis CI: Replace pep8 with flake8, update dependencies 

cclauss:
@bfalling PEP8 is deprecated so replace it with Flake8 and update some dependencies


jjjake:
Thanks @cclauss!

Issue #341, opened 2020-04-29
Connection reset by peer when uploading large file


Coloradohusky:
Uploading my 56GB ZIP file of jonesmekas.com, a website that was grabbed with grab-site, gives me the error `requests.exceptions.ConnectionError: (ProtocolError('Connection aborted.', error(104, 'Connection reset by peer')), 'https://s3.us.archive.org/jonasmekas.com-2020-04-28-952295c8/jonasmekas.com-2020-04-28-952295c8.zip')`. Using Ubuntu on Windows, attached file containing the full console output below and wrote the code I use for uploading to the Internet Archive below.

```
#!/bin/bash
ls -d */ | cut -f1 -d'/' > ./websites.txt
input="./websites.txt"
while IFS= read -r line
do
	echo "Zipping folder."
	zip -r -9 "$line"".zip" "./""$line"
	echo "Finished zipping folder."
	echo "Uploading zipped folder."
	./ia upload "$line" "$line"".zip" --metadata="mediatype:data" --metadata="creator:Coloradohusky" --metadata="description:warc created using grab-site" --metadata="subject:warc;archiveteam" --metadata="language:English"
	echo "Upload finished."
done < "$input"
rm websites.txt
```

[IAerror.txt](https://github.com/jjjake/internetarchive/files/4551874/IAerror.txt)

jjjake:
Hmm... Is this a consistent issue for you, or is it transient? Are you experiencing it with other files you try to upload?

Coloradohusky:
Happens with any large file, usually restarting the upload once or twice lets it upload successfully, but not this time

jjjake:
I'll try to replicate it, but have not been able to so far. Please let me know if you have any other details that might be helpful in troubleshooting this. Thanks for the report, and my apologies for the issues!

Coloradohusky:
Version 1.9.3, adding --debug gives me
```
jonasmekas.com-2020-04-28-952295c8:

 uploading jonasmekas.com-2020-04-28-952295c8.zip:   0%|                                   | 0/55428.0 [00:00<?, ?MiB/s]Endpoint:
 https://s3.us.archive.org/jonasmekas.com-2020-04-28-952295c8/jonasmekas.com-2020-04-28-952295c8.zip

HTTP Headers:
 Connection:close
 Accept-Encoding:gzip, deflate
 Accept:*/*
[...]

PR #372, opened 2020-09-15
Skip files which were uploaded completely, but the server did not return any response

Dobatymo:
In some cases the server closes the connection after the file was fully uploaded without sending a response. See my comment here https://github.com/jjjake/internetarchive/issues/341#issuecomment-689891649

This doesn't fix the issue of OP, but makes batch uploading of a large number of files easier. I am only handling the `RemoteDisconnected` exception, which is only raised when the client tries to read the http status line after sending the request.

In my testing the files were always uploaded completely in that case. So rerunning the batch upload with skipping with checksums enabled, it would skip these 'failed' failes correctly. So in case this error is encountered, it will show a informative message and continue with the next file instead of raises the exception and stopping the batch upload.


Dobatymo:
It seems flake8 fails because the `upload_item` function is "too complex".
Also the exception `BadStatusLine` used for Python2 is slightly more broad than `RemoteDisconnected` but should be ok.

soredake:
jjjake

Dobatymo:
Do you want to have another look at this PR? I have been using this patch for a year now and it improves uploading of large batches significantly for me.

The code doesn't do anything complex. Just some fine-grained exception handling to ignore some errors (raised incorrectly) by the server.

jjjake:
I'm not certain I like this as the default, but I think it could be handy as an option. I think it's important to raise an exception (or have a non-zero exit status code), if there's a chance the upload wasn't successful. What are your thoughts on that @Dobatymo?

Dobatymo:
I think raising an exception is not the way to go. This error is extremely common for me when uploading batches of large files. Sometimes every file encounters this error, but I have yet to find a file where this error occured and it was not actually fully uploaded. That's why I think retrying is not good also.

Of course I could return another exit code if this error was encountered for any of the files in the batch.

soredake:
What's the status of this?

Dobatymo:
@soredake The patch was working well 2-3 years ago. I haven't tested recently. New versions of requests or urllib3 might have changed.

Issue #377, opened 2020-10-03
ImportError: No module named collections_abc


2E0PGS:
```
sudo pip install internetarchive
```

```
$ ia
Traceback (most recent call last):
  File "/usr/local/bin/ia", line 6, in <module>
    from internetarchive.cli.ia import main
  File "/usr/local/lib/python2.7/dist-packages/internetarchive/__init__.py", line 45, in <module>
    from internetarchive.item import Item
  File "/usr/local/lib/python2.7/dist-packages/internetarchive/item.py", line 51, in <module>
    from internetarchive.utils import IdentifierListAsItems, get_md5, chunk_generator, \
  File "/usr/local/lib/python2.7/dist-packages/internetarchive/utils.py", line 38, in <module>
    from six.moves.collections_abc import Mapping
ImportError: No module named collections_abc
```

```
$ uname -a
Linux 4.19.0-11-amd64 #1 SMP Debian 4.19.146-1 (2020-09-17) x86_64 GNU/Linux
Distributor ID:	Debian
Description:	Debian GNU/Linux 10 (buster)
Release:	10
Codename:	buster
```

```
$ cat /etc/debian_version
10.6
```

Also tested previously with non root install. Did a uninstall before reinstalling.

The [binary](https://archive.org/services/docs/api/internetarchive/installation.html#binaries) does however work.

jjjake:
Thanks for reporting this. I believe updating the `six` module should resolve this for you:

```pip install --upgrade six```

Are you able to give that a try and report back? I will be increasing the minimum version requirement for `six` in the next release. Sorry for the trouble, and thanks again for reporting!

PR #378, opened 2020-10-06
Bump six version requirement

JustAnotherArchivist:
The module six.moves.collections_abc, added in commit 170c9639c0394e40d249f025e86b3ef79f4c4069, only exists since six version 1.13.0.


jjjake:
Thanks a lot, @JustAnotherArchivist!

JustAnotherArchivist:
Hadn't seen the issue before, but this fixes #377.

Issue #203, opened 2017-07-21
Check for valid identifier before uploading?


TheCoinCollector:
I'm still relatively new to using this module, so I wonder if this is this way for a reason.

When attempting an upload with -spreadsheet with what turns out to be an invalid identifier, it waits until the full file gets uploaded and _then_ fails with an error message.  This just happened on a pretty large video file.  Is there any reason it can't check the identifier before beginning the upload?

Thanks!

jjjake:
How is the upload failing exactly @TheCoinCollector? What is the error message? Can you also send me the output of `ia --version`?

The Python library that `internetarchive` uses to make HTTP requests, Requests, does not have 100-continue support and our S3 API depends on that for raising certain errors without waiting for all of the data to be sent.

We might be able to work around this client-side, depending on the error.

TheCoinCollector:
I'm on 1.5.0 - downgraded for now to get around the separate broken pipe error.

The identifier error I mentioned here was no big deal - it was longer than 100 characters.  I just thought it was strange ia did the full upload before giving me the error, but if it's because of a technical hurdle that's fine.

PR #381, opened 2020-11-21
Validate identifiers in spreadsheet before uploading file

smrohrer:
If a spreadsheet for bulk upload includes an invalid identifier, the CLI would upload the entire file before reporting the error, and the entire file would have to be uploaded again after fixing the identifier.

This was previously noted in #203.

The CLI will now validate the identifier on the client side before uploading, and will terminate the program if an invalid identifier is encountered.

### DIFF
--- a/internetarchive/cli/ia_upload.py
+++ b/internetarchive/cli/ia_upload.py
@@ -125,6 +125,21 @@ def _upload_files(item, files, upload_kwargs, prev_identifier=None, archive_sess
     return responses
 
 
+def _validate_identifier(identifier):
+    if not identifier:
+        print('error: no identifier column on spreadsheet!')
+        sys.exit(1)
+    try:
+        validate_ia_identifier(identifier)
+    except AssertionError:
+        print('error: identifier "{}" is invalid. '.format(identifier) +
+            '<identifier> should be between 3 and 80 characters in length, and '
+            'can only contain alphanumeric characters, periods ".", '
+            'underscores "_", or dashes "-". However, <identifier> cannot begin '
+            'with periods, underscores, or dashes.')
+        sys.exit(1)
+
+
 def main(argv, session):
     if six.PY2:
         args = docopt(__doc__.encode('utf-8'), argv=argv)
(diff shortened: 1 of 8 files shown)


internetarchive/cli/ia_upload.py
```diff
@@ -265,6 +265,15 @@ def main(argv, session):
                 if not identifier:
                     print('error: no identifier column on spreadsheet!')
                     sys.exit(1)
+                try:
+                    validate_ia_identifier(identifier)
+                except:
```
cclauss, line 270: PEP8 and https://realpython.com/the-most-diabolical-python-antipattern document why bare exceptions should be avoided.

smrohrer: Thanks for pointing that out

jjjake:
Thanks for the contribution!

I think this should be an option (off by default) in case the rules for a valid identifier shift, and I think the option should work for both spreadsheet mode and regular uploads. I also think that the `_validate_identifier` function should be moved to `internetarchive.utils.validate_identifier`, and implemented in `internetarchive.upload_file` and `internetarchive.upload` (and turned on in `ia_upload.py` in the Item.upload method calls). Perhaps it should be named something like `validate_s3_identifier()` as well, because there are different rules for what is considered a valid identifier at item creation vs. modifying metadata, for example.

Please let me know if you have any questions. I think this would be a helpful feature to help deal with the `requests` module lack of support for 100-continue (this is what requires the whole file to be uploaded before an error is returned by IA-S3).

smrohrer:
Thank you Jake for the feedback! I will push up updates shortly.

To be clear, do you want the CLI to not validate identifiers by default, both for spreadsheet and regular uploads? Currently for regular uploads via the CLI, the identifier is always validated as part of the arg scheme validation.

Thanks!

jjjake:
Hey @smrohrer, sorry for the confusion and thank you for the contribution. I deleted my comment after posting because I realized I needed to take a closer look at the current code. I forgot that we're already validating for non-spreadsheet-mode uploads by default.

With that in mind, your changes look great. I would just suggest that we _do_ validate by default to be consistent with the existing behavior. I don't think we need a command-line option to toggle the validation in this case, since it's been the default for years without issue.

TL;DR, your changes look great. Let's just remove the `--validate-identifier` option, and validate identifier's by default. Let me know if you have any questions or thoughts on this. Thanks again!

smrohrer:
Thank you Jake for reviewing, glad I can help make the tool a little better!

jjjake:
Looks great, @smrohrer, thank _you_! Thanks for writing tests as well! :)

PR #383, opened 2020-11-23
feat: don't retry if spam detected

### LINKED ISSUE #298: Add indication for 503 Slow Down Spam error.
Some items fail to upload because they have been marked as spam, but the script treats this situation as a simple server overload, so it retries indefinitely. Maybe it would be possible to detect this by verifying whether the request queue is full, or through whatever means allows the browser uploader to give a unique message when an upload has been marked as spam.

midgleyc:
Fixes #298 . I know it's already closed as an issue for archive.org, but the problem is still happening, and this change helps.

Occasionally a mirrored item can be marked as spam. In this case there's not much point retrying as it'll be marked as spam again. This changes ia to abort early if that happens.

I didn't change `get_s3_xml_text` to include the RequestId, so the error is just:
```
 error uploading cRRrauRtugc.description: Please reduce your request rate. - Your upload of youtube-cRRrauRtugc from username ***@gmail.com appears to be spam. If you believe this is a mistake, contact info@archive.org and include this entire message in your email.
```

If the RequestId helps (or if "this entire message" should be something different) I can look at adding it.

### DIFF
--- a/internetarchive/item.py
+++ b/internetarchive/item.py
@@ -1059,6 +1059,9 @@ class Item(BaseItem):
                                                  stream=True,
                                                  **request_kwargs)
                     if (response.status_code == 503) and (retries > 0):
+                        if b'appears to be spam' in response.content:
+                            log.info('detected as spam, upload failed')
+                            break
                         log.info(error_msg)
                         if verbose:
                             print(' warning: {0}'.format(error_msg), file=sys.stderr)


XXLuigiMario:
This would be really nice to have. I run into this issue fairly often as some bit of metadata flags the upload and there's nothing to be done at that point.

nemobis:
I have the opposite issue: since the error 503 is temporary, I'd like the upload to be retried. Instead, I get an exception, where it would be enough to retry one time with a generous wait time, possibly multiple minutes. 

Just giving up and throwing an exception makes it more complicated to make sure all my files were uploaded correctly, because I have to implement all the completion checks and retrying myself.

```
  File "/usr/local/lib/python3.10/site-packages/internetarchive/item.py", line 1091, in upload_file
    response.raise_for_status()  
  File "/usr/lib/python3.10/site-packages/requests/models.py", line 960, in raise_for_status       
    raise HTTPError(http_error_msg, response=self)                                                                    
requests.exceptions.HTTPError: 503 Server Error: Slow Down for url: https://s3.us.archive.org/ ...                                                                                                                                                                                                                                                         
During handling of the above exception, another exception occurred:
                                                                                                                      
...                        
  File "/usr/local/lib/python3.10/site-packages/internetarchive/api.py", line 289, in upload
    return item.upload(files,                                                                                         
  File "/usr/local/lib/python3.10/site-packages/internetarchive/item.py", line 1260, in upload
    resp = self.upload_file(body,
  File "/usr/local/lib/python3.10/site-packages/internetarchive/item.py", line 1117, in upload_file
    raise type(exc)(error_msg, response=exc.response, request=exc.request)
requests.exceptions.HTTPError:  error uploading , Please reduce your request rate. - Your upload of ... appears to be spam. If you believe this is a mistake, contact info@archive.org and include this entire message in your email.

During handling of the above exception, another exception occurred:
```

midgleyc:
I would appreciate it if you could add that, as I'm not too familiar with this codebase. Thanks!

internetarchive/item.py
```diff
@@ -1059,6 +1059,9 @@ def _build_request():
                                                  stream=True,
                                                  **request_kwargs)
                     if (response.status_code == 503) and (retries > 0):
+                        if b'appears to be spam' in response.content:
```
cclauss, line 1062:
```suggestion
                        if 'appears to be spam' in response.text:
```

PR #394, opened 2021-03-16
Fix #393 by setting empty response status code

### LINKED ISSUE #393: ia upload --spreadsheet --checksum fails. TypeError: '<=' not supported between instances of 'int' and 'NoneType'
Uploading with both --spreadsheet and --checksum currently fails:
```
  File "/home/joel/.local/bin/ia", line 10, in <module>
    sys.exit(main())
  File "/home/joel/.local/lib/python3.7/site-packages/internetarchive/cli/ia.py", line 170, in main
  File "/home/joel/.local/lib/python3.7/site-packages/internetarchive/cli/ia_upload.py", line 290, in main
    if (not _r) or (not _r.ok):
  File "/home/joel/.local/lib/python3.7/site-packages/requests/models.py", line 679, in __bool__
    return self.ok
  File "/home/joel/.local/lib/python3.7/site-packages/requests/models.py", line 705, in ok
    self.raise_for_status()
  File "/home/joel/.local/lib/python3.7/site-packages/requests/models.py", line 936, in raise_for_status
    if 400 <= self.status_code < 500:
TypeError: '<=' not supported between instances of 'int' and 'NoneType'
```

The reason for this is here:
https://github.com/jjjake/internetarchive/blob/55431a5f028dfab022ad0f01626e89a5309ad7f7/internetarchive/item.py#L992-L995

If `checksum=True`, `item.upload` returns a default `Response` object if an identical file is found on the server. The bulk upload code in `ia_upload.py` calls `ok` on each response received. However, since a default `Response` object has `status_code` initialized to `None`, it gets tripped up in `raise_for_status`.

A band-aid solution would be to set `status_code` to 200 before returning the `Response` object.

XXLuigiMario:
Takes care of #393 by setting `status_code`, allowing `r.ok` to succeed.

### DIFF
--- a/internetarchive/item.py
+++ b/internetarchive/item.py
@@ -990,9 +990,12 @@ class Item(BaseItem):
                     body.close()
                     os.remove(filename)
                 # Return an empty response object if checksums match.
+                # status_code must be set. See issue #393
                 # TODO: Is there a better way to handle this?
                 body.close()
-                return Response()
+                r = Response()
+                r.status_code = 200
+                return r
 
         # require the Content-MD5 header when delete is True.
         if verify or delete:


XXLuigiMario:
Ouch. It appears that the checksum test relies on `status_code` being `None` to tell if a file was skipped.
In that case, it might be wise to follow the TODO's advice and handle this differently.

JustAnotherArchivist:
Obsoleted by #478, although a proper solution for that TODO would still be nice, but that's for another day.

PR #396, opened 2021-03-17
item.py: Simplify logic for readability

cclauss:


jjjake:
Thanks once again, @cclauss!

PR #398, opened 2021-04-17
Simplify the logic of identifier_available, making it more idiomatic and readable.

maxz:
When looking how you handled to check for available identifiers, I came across this rather unidiomatic statement which could be simplified by reversing the unnecessary negation, since the check_identifier.php returns `<result type="success" code="available">` ([example](https://archive.org/services/check_identifier.php?identifier=quux)) if the identifier is available and the previously relied upon `<result type="success" code="not_available">` ([example](https://archive.org/services/check_identifier.php?identifier=foobar)) if it is unavailable.

It should also correspond better with the actual availability, in case the codes ever get expanded in the future and currently maps more directly to the method's name.

By requesting the [seemingly more modern JSON](https://archive.org/help/json.php) instead of XML data, the somewhat hairy availability code extraction could be further simplified from a DOM tree traversal into a dictionary lookup.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -12,6 +12,7 @@ Contributors
 ------------
 
 - Bryce Drennan <internetarchive@brycedrennan.com>
+- Max Zettlmeißl <max@zettlmeissl.de>
 
 Patches and Suggestions
 -----------------------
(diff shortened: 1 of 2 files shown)


jjjake:
I didn't even realize check_identifier.php could output JSON. This change sounds great. I'll merge it after I take a closer look. Thanks for the contribution, @maxz!

PR #401, opened 2021-04-23
Fix return type for modify_metadata

DuncanDHall:
The correct rtype is requests.Response, not dict


jjjake:
Thanks, @DuncanDHall!

Issue #399, opened 2021-04-21
Config file location defaults to `~/.ia` instead of `~/.config/ia.ini`


JustAnotherArchivist:
I discovered today that my PR #324 inadvertently introduced a bug and changed the default config file path from `~/.config/ia.ini` to `~/.ia`. I didn't realise at the time that `write_config_file` uses `parse_config_file` to get the path to write to. If neither file exists, `parse_config_file` should return `~/.config/ia.ini`.

Possibly, this should only be the case if `~/.config` exists. Personally though, I think that the directory should just be created with `os.makedirs` if it doesn't exist as it has been the standard location for config files of any software for oh-so-many years now and home dir pollution is horrible. As that would be a deliberate functional change though, I'd like to hear @jjjake's thoughts on that first. (Will create a PR afterwards.)

jjjake:
Thanks for catching, I'm rolling #324 back.

Issue #385, opened 2020-12-06
ItemLocateError broken on Python 2


JustAnotherArchivist:
internetarchive claims Python 2 compatibility, but `ItemLocateError` (introduced by deb0d91a and first released in 1.9.2) makes use of `super()`, which is only available in this argument-less form on Python 3.

```
> python2 -c 'from internetarchive.exceptions import ItemLocateError; ItemLocateError()'
TypeError: super() takes at least 1 argument (0 given)
```

I haven't checked whether there are any other issues with Python 2 as I uninstalled that after it reached EOL at the beginning of this year.

Credit for finding this goes to @jodizzle.

JustAnotherArchivist:
Fixed by dropping Python 2 support per #435.

PR #405, opened 2021-05-05
Implement XDG Base Directory specification

### LINKED ISSUE #400: `XDG_CONFIG_HOME` is not respected
The primary config file location is currently hardcoded as `~/.config/ia.ini`. Instead, it should use the `XDG_CONFIG_HOME` environment variable if set, defaulting to `${HOME}/.config` ([specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html)).

I'll fix this together with #399 but wanted to file an issue anyway so it isn't forgotten.

JustAnotherArchivist:
`ia` will attempt the following paths in this order, using the first existing file:

- `${XDG_CONFIG_HOME}/internetarchive/ia.ini` (with `XDG_CONFIG_HOME` defaulting to `${HOME}/.config` if not set or invalid)
- `~/.config/ia.ini`
- `~/.ia`

When none of these paths exist on creating a config file (`ia configure`), the first one is used and directories are created as necessary.

Closes #400

### DIFF
--- a/internetarchive/api.py
+++ b/internetarchive/api.py
@@ -572,8 +572,8 @@ def configure(username=None, password=None, config_file=None, host=None):
     """
     username = input('Email address: ') if not username else username
     password = getpass('Password: ') if not password else password
-    config_file_path = config_module.write_config_file(
-        username, password, config_file, host)
+    auth_config = config_module.get_auth_config(username, password, host)
+    config_file_path = config_module.write_config_file(auth_config, config_file)
     return config_file_path
 
 
--- a/internetarchive/config.py
+++ b/internetarchive/config.py
@@ -72,9 +72,8 @@ def get_auth_config(email, password, host=None):
     return auth_config
 
 
-def write_config_file(username, password, config_file=None, host=None):
+def write_config_file(auth_config, config_file=None):
     config_file, is_xdg, config = parse_config_file(config_file)
-    auth_config = get_auth_config(username, password, host)
 
     # S3 Keys.
     access = auth_config.get('s3', {}).get('access')
@@ -88,7 +87,7 @@ def write_config_file(username, password, config_file=None, host=None):
     config.set('cookies', 'logged-in-sig', cookies.get('logged-in-sig'))
 
     # General.
-    screenname = auth_config['general']['screenname']
+    screenname = auth_config.get('general', {}).get('screenname')
     config.set('general', 'screenname', screenname)
 
     # Create directory if needed.
--- a/internetarchive/config.py
+++ b/internetarchive/config.py
@@ -73,7 +73,7 @@ def get_auth_config(email, password, host=None):
 
 
 def write_config_file(username, password, config_file=None, host=None):
-    config_file, config = parse_config_file(config_file)
+    config_file, is_xdg, config = parse_config_file(config_file)
     auth_config = get_auth_config(username, password, host)
 
     # S3 Keys.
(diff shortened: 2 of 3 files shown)


JustAnotherArchivist:
A note on `os.environ['HOME']` vs `os.expanduser('~')`: the XDG specification requires using `$HOME`. `os.expanduser` falls back to using the password database or the root directory if `$HOME` doesn't exist and is therefore not fit for purpose. In practice, this shouldn't matter as POSIX *requires* the presence of that environment variable, but I feel like it was worth mentioning anyway.

JustAnotherArchivist:
For some reason, `$HOME` apparently isn't set in Travis CI. Lovely...
Not sure what the reason for that is. Smells like a bug on Travis's side.

I've successfully tested this code locally with a POSIX-compliant environment in all scenarios I could think of.

JustAnotherArchivist:
Rethinking this, I suppose `ia` should also work on Windows, where `$HOME` doesn't exist as far as I know. Perhaps using `~` is better after all; on any properly POSIX-compliant system, it will still behave according to the specification since `expanduser` uses `$HOME` if set. @jjjake, any thoughts on this?

JustAnotherArchivist:
I made that change from `$HOME` to `~` as I can't really think of a good reason not to do that. The specification doesn't say anything about what should happen if `$HOME` isn't present, presumably because it assumes a POSIX-compliant system (without mentioning it). And on such systems, the new code will still behave according to the specification because Python's expansion uses `$HOME` if present on Unix and only falls back to the password database if not; on Windows, `$HOME` is useless anyway.

I would still like to figure out why the build failed though. I wrote to Travis CI, and they confirmed that `$HOME` works correctly in an independent test build. I'll probably create a separate PR to play with that.

Also, I still want to add some tests here to ensure that a #399 situation can't happen again. So this isn't ready for merging yet.

JustAnotherArchivist:
Tests are succeeding except on Python 2.7 because `tempfile.TemporaryDirectory` doesn't exist there. I could try to reimplement that, but Py2 is EOL and ia is already broken on it (#385), so I'm not sure that's worth the effort. @jjjake, please advise.

JustAnotherArchivist:
I took a quick look at the implementation of `tempfile.TemporaryDirectory`, and it's more complex than I thought. If dropping Py2 support is not yet an option, skipping these tests on Py2 would be another possibility.

JustAnotherArchivist:
Alright, I understand. To be clear, it's only the tests that are held back by this; the actual code should not be Py3-specific in any way.

I'll try to write a minimal crutch for Py2 that emulates the key part of `TemporaryDirectory`. I have no desire to reimplement its [complex error handling](https://github.com/python/cpython/blob/b4884083aed9bfec22c4da0e0223ad6a04a4bef2/Lib/tempfile.py#L766) though – a lot of things in that area changed between Py2 and Py3 and I haven't written code for Py2 in years... But it should be good enough to get the test suite running on CI at least.

JustAnotherArchivist:
I realised today that `write_config_file` was broken on Python 2 for the same reason as the tests were previously (lack of `exist_ok` kwarg), but because there were no tests for that method at all, it didn't trip anything. I refactored that code slightly to decouple the IA login from the config file writing and added tests for all of that as well.

This should now be ready for review and merging. :-)

jjjake:
Thanks again, @JustAnotherArchivist!

PR #406, opened 2021-05-08
Fix `--config-file` being ignored when `ia configure` is used with `--username` and `--password`

JustAnotherArchivist:


jjjake:
Thank you!

PR #411, opened 2021-06-05
Test Travis CI HOME environment variable failure

JustAnotherArchivist:
Please disregard this PR, and apologies for the noise. I am trying to debug the weird missing `HOME` environment variable that showed up in one of the builds for #405: https://travis-ci.com/github/jjjake/internetarchive/builds/224924797

Travis CI's support was unable to reproduce it and I've never seen it in another project either, so working backwards from a failing build is the only reasonable way it seems.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,13 +3,5 @@ dist: xenial
 language: python
 install: pip install tox-travis
 python:
-    - "2.7"
-    - "3.6"
-    - "3.7"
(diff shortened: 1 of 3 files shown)


JustAnotherArchivist:
This turns out to be an issue in tox, which does not pass `HOME` to subprocesses by default. Filed upstream: https://github.com/tox-dev/tox/issues/2083

PR #418, opened 2021-07-19
Output a more helpful error message on malformed bulk upload CSV data.

maxz:
Malformed metadata keys in a bulk upload CSV file previously lead to a somewhat obscure ConnectionResetError.  Now the offending key is named in an error message which should greatly help users who encounter this problem.

The lxml module is a required dependency since the standard library's XML module does not properly validate tag names.
I will also use the lxml module for the multipart-upload feature I'm currently working on, so the dependency won't just be introduced for this single check.

Closes: #417


jjjake:
Are we able to do this without adding `lxml` as a requirement? I have not had good experiences, especially with building universal binaries, with adding C based libraries as requirements.

Ease of installation, use of binaries, and compatibility is a high priority.

Otherwise this looks great, thank you!

maxz:
Not with the standard library XML parser.
But a regex would always work I guess.
It was among the first possible solutions I took a look at.

I will change to a regex based solution as soon as I have time.

maxz:
Done. I just squashed all the commits into one. The addition of the `is_valid_metadata_key` function and introducing its usage during the CSV bulk upload are not separate commits anymore, but this prevents the introduction and removal of the `lxml` module from showing up in the commits.

jjjake:
Looks great, thank you @maxz!

Issue #413, opened 2021-06-12
Creating an item with `ia copy` results in unexpected metadata


JustAnotherArchivist:
Last night, I created https://archive.org/details/stackexchange_20210607 using this command (description removed for brevity):

    ia copy stackexchange/3dprinting.meta.stackexchange.com.7z stackexchange_20210607/3dprinting.meta.stackexchange.com.7z --metadata=mediatype:data --metadata=collection:opensource_media --metadata=date:2021-06-07 '--metadata=title:Stack Exchange Data Dump 2021-06-07' '--metadata=subject:Stack Exchange Data Dump' '--metadata=contributor:Stack Exchange Community' '--metadata=creator:Stack Exchange, Inc.' --metadata=description:snip

For comparison, here's the previous item: <https://archive.org/details/stackexchange_20210301>. The only difference between the two items is that the March one was created with `ia upload` from a local copy and the June one with the `ia copy` command above. Except for the changed dates, the `--metadata` options were exactly the same. As such, I would expect the items' metadata to be virtually identical as well. (The June item was also created with an older `ia` version than the March one, but I don't see any relevant changes in that part of the code since long before 1.9.3, so I'm not sure that's in any way relevant.)

Here are some notable oddities in the June item:

* It has two `mediatype` fields: `<mediatype>texts</mediatype>` and `<mediatype>data</mediatype>`
* It is part of the opensource collection: `<collection>opensource</collection>`
* It has two metadata fields that I've never seen before and one of which (`identifier-access`) is undocumented: `<identifier-access>http://archive.org/details/stackexchange_20210607</identifier-access>` and `<identifier-ark>ark:/13960/t5kb4rz3h</identifier-ark>`
* It has a `creator` field: `<creator>StackExchange</creator>`

None of these are present on the source item.

I suspect that all of this is a bug in the backend rather than `ia copy`, but since the latter is where it surfaces, I figured I'd file an issue here anyway. The texts mediatype and opensource collection values are the defaults when nothing else is specified, so apparently those get added even when others are present in the copy request. The two `identifier-*` ones are probably harmless but still odd. The creator field is the weirdest of all because it doesn't match *anything* as far as I can see; that specific spelling without a space and capitals doesn't seem to appear anywhere...

maxz:
> Here are some notable oddities in the June item:
> 
> * It has two `mediatype` fields: `<mediatype>texts</mediatype>` and `<mediatype>data</mediatype>`
[...]

PR #428, opened 2021-08-26
Give precedence to the most recent metadata during copy operations.

maxz:
When specifying metadata during a copy operation, any new metadata specified on the command line would be added to the item in addition to the fully copied original metadata which could lead to very bizarre results and even metadata conflicts unresolvable by regular users.

Now only the parts of the old metadata which don't conflict with newer metadata are copied.
File metadata is now also copied over.

It may or may not be desirable to have a more fine grained merge on some metadata keys for which it might make sense (e. g. the subjects), but I did not want to make that call at this time.

The new switch `--replace-metadata` drops all metadata from the original item and only uses metadata specified as argument to `ia` (or the default value).
The new switch `--ignore-file-metadata` prevents the copying of any file metadata.

Closes: #413

### DIFF
--- a/internetarchive/cli/ia_copy.py
+++ b/internetarchive/cli/ia_copy.py
@@ -114,6 +114,12 @@ def main(argv, session, cmd='copy'):
         args['--metadata'] = merge_dictionaries(SRC_ITEM.metadata,
                                                 args['--metadata'])
 
+    # File metadata is copied by default but can be dropped.
+    if args['--ignore-file-metadata']:
+        file_metadata = None
+    else:
+        file_metadata = SRC_FILE.metadata
+
     # Add keep-old-version by default.
     if 'x-archive-keep-old-version' not in args['--header']:
         args['--header']['x-archive-keep-old-version'] = '1'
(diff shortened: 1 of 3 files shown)


internetarchive/cli/ia_copy.py
```diff
@@ -97,7 +102,16 @@ def main(argv, session, cmd='copy'):
         sys.exit(1)
 
     args['--header']['x-amz-copy-source'] = '/{}'.format(parse.quote(src_path))
-    args['--header']['x-amz-metadata-directive'] = 'COPY'
+    # Copy the old metadata verbatim if no additional metadata is supplied,
+    # else combine the old and the new metadata in a sensible manner.
+    args['--header']['x-amz-metadata-directive'] = ('REPLACE' if (args['--metadata']
```
jjjake, line 107:
```suggestion
    args['--header']['x-amz-metadata-directive'] = ('REPLACE' if (args['--metadata']
```

jjjake, line 107: Could you please use multi-line statements in cases like these? Thanks again!

jjjake:
This sounds great, thank you! I left a comment about multi-line statements, otherwise looks great!

internetarchive/cli/ia_copy.py
maxz: Is the latest commit what you had in mind?

The duplication of `args['--header']['x-amz-metadata-directive']` seems like the most readable form to me.

Issue #337, opened 2020-04-19
The --retries flag isn't working for ia download


jakewilliami:
Using `ia` version 1.9.3.  I am running, for example
```
ia download bangla-tutoria-book 'Advanced Excel 2013 Bangla Book .gif' --retries 10
```
However, this is just giving me the download usage help message (as if the syntax isn't correct).  I have checked the docs and source code and it seems that I am using the `--retries` flag correctly.  I downloaded internetarchive via `pip3` on my Apple machine.

maxz:
Your comment just helped me to solve this, @JustAnotherArchivist.
I looked at all parts of this except for the usage listing, because I wrongly assumed that this would not matter.

The documentation for `docopt` is really bad by the way. When I began troubleshooting this I tried to gain the required knowledge by reading the documentation, but it is almost non-existent.

PR #429, opened 2021-08-28
Fix the handling of the retries argument for downloads.

maxz:
One of the usage directives for the download subcomponent of `ia` was specified incorrectly and therefore supposedly valid combinations of command line arguments could not be parsed correctly.

This now works correctly.

Additionally I fixed a minor "beauty issue".

Closes: #337


JustAnotherArchivist:
I'm almost certain that the reason behind the previous usage line is that most options of `ia download` can't be used with `--stdout` (`--retries` and `--parameters` being the exceptions). This new code accepts all options but then silently ignores nearly all of them, which in my eyes isn't great and may lead to confusion. For example, *'why doesn't `ia download $identifier $file --stdout --verbose` tell me what's going on via stderr‽'* (Answer: `--verbose` is a no-op with `--stdout` and would write to stdout anyway.)

maxz:
`[option]...` is a wildcard for all arguments listed in the options section except for those which appear in any usage pattern. So in the unpatched case, any argument except for `--stdout`,  `--retries`, `--itemlist`, `--search` and `--help`.

There does not seem to be a working alternative to this patch in combination with `docopt` and it certainly is better to have all options that are supposed to work working and some options silently ignored than to have the usage message pop up when something that is supposed to work is entered.
Listing `--parameters` and `--retries` in all patters, which would also work, is not a very appealing alternative.

Merely outputting the usage note in cases where e. g. `--verbose` is not allowed would not help the user either.

`--verbose` could receive a special treatment in combination with `--stdout`.
Although there are many other cases where options are silently ignored in `ia` e. g. right within `ia_download.py` many options combined with `--dry-run`.
Silently swallowing some arguments is rather common with shell programs.

I just went over the `docopt` documentation once more and tested some things regarding the command grammar, but nothing except the current solution results in the desired behaviour
I also took a look at `docopt-ng` but although it was forked to improve the maintenance situation of `docopt`, it sadly is not maintained any better and as far as I could tell by looking at the history, the issues and the pull request, this particular issue is not fixed there either. The grammar of that rule would actually have still worked in 2013, until a breaking change was intentionally introduced with version 0.6 of `docopt`.

jjjake:
Thanks @maxz! Can you please keep the `ignore_history_dir = True if not args['--download-history'] else False` bit in though? I prefer the clarity. Thanks again!

maxz:
I rebased my branch, dropping every commit which had changes related to `ignore_history_dir`.

PR #430, opened 2021-08-30
Documentation: Specify the maximum allowed length of identifiers.

maxz:
The other Internet Archive documentation specifies the maximum length
as 100 characters, but actually it is 101 characters as you can test
by creating such an item and trying to create one which is greater
than 101 characters.

This is probably an off-by-one error in the Internet Archive program
which checks identifiers.

See: https://archive.org/services/docs/api/metadata-schema/index.html

### DIFF
--- a/docs/source/metadata.rst
+++ b/docs/source/metadata.rst
@@ -16,7 +16,7 @@ Please refer to the XML Naming Rules section `here <https://www.w3schools.com/xm
 Archive.org Identifiers
 -----------------------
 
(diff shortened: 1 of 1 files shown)


jjjake:
There are various limits to archive.org identifier length, and 100 is what we recommend as the hard limit.

maxz:
What do you mean by various limit? The limit for regular users seems to be 101 characters.
Do you mean that different access methods can have different limits (e. g. administrative accounts being able to create identifiers > 101 characters)?

I can understand if you would rather want to see 100 instead of 101 characters in the documentation, but the current documentation says `While there are no official limits` which from my regular user perspective does not seem to be true, especially since the other documentation at https://archive.org/services/docs/api/metadata-schema/index.html explicitly mentions 100 characters. (`The maximum length of an identifier is 100 characters, but we generally recommend that identifiers be between 5 and 80 characters in length.`)

PR #441, opened 2021-10-16
Replace two deprecated names by their newer variants.

maxz:
While running the `internetarchive` test suit to check a different set of changes, I encountered deprecation warnings by `urllib3` and replaced the deprecated names by their newer variants.

The option to use the old names will be fully removed in version 2 of `urllib3`.
It should be noted that the support for Python 2 will also be dropped at that time.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,10 +39,11 @@ setup(
         'backports.csv < 1.07;python_version<"3.4"',
         'docopt>=0.6.0,<0.7.0',
         'jsonpatch>=0.4',
-        'requests>=2.9.1,<3.0.0',
+        'requests>=2.25.0,<3.0.0',
         'schema>=0.4.0',
         'six>=1.13.0,<2.0.0',
         'tqdm>=4.0.0',
+        'urllib3>=1.26.0',
     ],
     classifiers=[
         'Development Status :: 5 - Production/Stable',


JustAnotherArchivist (changes requested):
This requires `urllib3>=1.26.0` ([relevant commit](https://github.com/urllib3/urllib3/commit/382ab32f23795c44faae83b4e8b18a16fb605a0a)). However, we don't currently list `urllib3` as a dependency at all in `setup.py` (and `pex-requirements.txt`), which makes sense because that package isn't imported/used directly. `requests` depends on `urllib3>=1.21.1,<1.27`; the minimum version listed in our `setup.py` doesn't even properly depend on it because it was vendored at the time...

In other words, this change, without modifications to `setup.py`, would brick installations with old versions of `requests` or `urllib3` where only `internetarchive` gets upgraded since `pip` would think that no other changes are necessary. Likely there'd also be no obvious indication of what's wrong, i.e. that those packages need to be upgraded.

I guess adding `urllib3>=1.26.0` as an explicit dependency in `setup.py` is the least messy way out of that. Modern `pip` versions should then realise whether `requests` needs to be upgraded as well on a `pip install --upgrade internetarchive` to satisfy that requirement. Might be good to test this though.
A bump of the version requirement for `requests` may also be necessary to handle installations with ancient versions of `requests` that still vendored `urllib3`; that would be `requests` version 2.16.0 or newer. A possibly better one would be to bump it to the first version that supported `urllib3==1.26.0`, and that's 2.25.0 (even Debian stable has that...).

This is not a hypothetical scenario, by the way. Trying to prevent another #377.

maxz:
Thank you for the version analysis.

So `urllib3 1.26.0` is the minimum version required to handle the new names and `requests 2.25.0` is most likely the version of `requests` which will pose the least problems with that `urllib3` change.

The choice would then be whether to skip the change to the names for now or whether to bump those dependency versions. I guess I will just bump the versions for now and @jjjake can then decide whether or not to incorporate the changes at this time.

maxz:
`requests` depends on `urllib3>=1.21.1`, since at least version 2.25.0.
Adding an explicit dependency on `urllib>=1.26.0` to `internetarchive` would therefore be required.
It is a bit strange if we required a higher version, but without the higher version, my changes won't work and we will run into the same problem in the future.

`requests 2.25.x` is also the highest version to still support Python 3.5 so it would fit into the range of Python versions we currently support.

It would be wise to at least strictly depend on `urllib3<2.0.0` if we don't make those changes, else version 2.0.0 would break `internetarchive`.

JustAnotherArchivist (approved):

jjjake:
Thanks for being proactive on this! Just to confirm, this will still work on PY2? I would like to continue to support PY2 as long as is reasonable. Keeping things easy for those with less technical ability is a high priority, and unfortunately there's still a lot of PY2 in the wild. I'm really looking forward to moving beyond PY2, but I don't think we're quite there yet.

If this works in PY2, I'll go ahead and merge. If not, I'd prefer to hold out until at least urllib3 version 2.0.0 is released. Thanks again!

PR #446, opened 2021-11-23
Restore the CSV field validation.

maxz:
These commits fix the previous Python 2 incompatibility of `is_valid_metadata_key` by using a `re.match` expression which is anchored to have the same effect as the `re.fullmatch` expression.


jjjake:
Thanks @maxz, I appreciate it!

PR #449, opened 2021-11-24
v2.2.0

JustAnotherArchivist:
Fixes #448

Issue #435, opened 2021-09-18
Consider dropping Python 2 support


JustAnotherArchivist:
It's been almost two years since Python 2 has reached end-of-life. Some things already broke, and this will only get worse over time. Also, the code is significantly more complex (everything involving six; see also my struggles on #405).

I suggest dropping Python 2 support and removing the accompanying code cruft.

@jjjake has previously said in #405 that he 'would like to support [Python 2] as long as we're able to'. As such, I don't expect this to be acted on immediately. But I think it is inevitable as dependencies have begun to drop Py2 support, slowly making it harder to keep the support here.

Relevant issues so far (in no particular order):

* `ItemLocate` error causes crash under Python 2: #385
* contextlib2, an upstream dependency of schema, no longer supports Python 2; by extension, the self-contained binary is now broken: https://github.com/jjjake/internetarchive/issues/420#issuecomment-922371186
* `ia tasks` on Python 2 can't work reliably (without API changes) due to the lack of appropriate decoding error handlers: #359 and #360

As of right now, the contextlib2 development breaks new ia installations entirely on Python 2 unless you manually install an older version of that package. It could be pinned, but that's just delaying the inevitable.

Further things to look forward to:

* pip has removed Python 2 support as of version 21.0: https://github.com/pypa/pip/issues/6148 https://github.com/pypa/pip/pull/9189
* Requests maintainers have previously stated that they will drop support once pip does: https://github.com/psf/requests/issues/5211

maxz:
Since many of the libraries that we rely on will soon be dropping their Python 2 support, I'm generally in favour of also dropping the Python 2 support for `internetarchive`.

In the past I've had to deal with environments where I could only easily run Python 2, since running Python 3 would have required that I cross compile it myself and somehow get it to permanently reside in the memory of those rather special systems, so there are certain use cases I see for which it would be good to keep Python 2 support.

But the only realistic option to keep the Python 2 support would be to pin the latest Python 2 supporting versions of all our dependencies, which is pretty bad from a security point of view.
Backporting the newer versions of all our dependencies to Python 2 is just not a realistic option.

Therefore it is most likely the best option to just drop the Python 2 support.

I don't see any reason to do so right away, but as soon as the first major dependency drops the support, we pretty much also have to do so. Maybe we should prepare for that time by stating in all important places, that the Python 2 support will be dropped to give the users time to prepare (and try to find out the rough time when this will take place).

cclauss:
[...]

PR #450, opened 2021-11-24
Synchronise tested Python versions in tox.ini and .github/workflows/tox.yml

JustAnotherArchivist:
#437 didn't use the same Python version in the GitHub Actions config as in the tox config. As a result, Python 2 and 3.5 aren't tested automatically on PRs etc.

This synchronises the two lists. Namely, it restores CPython 2 and 3.5 testing on GitHub Actions and adds PyPy to the tox configuration.

As expected (#435), installations on the versions 2.7 and 3.5 do not currently work due to dependencies no longer supporting those EOL versions, so those jobs will fail until they are fixed in further PRs. Unfortunately, GitHub Actions still doesn't support allowable failures (actions/toolkit#399), so this will show red crosses everywhere. I think that's preferable to not running the tests at all though, otherwise the code will likely break even further under those ancient Python versions, which makes a final compatibility release as desired by @jjjake in #435 harder.

NB, I made this commit on top of the v2.2.0 release, which isn't on master yet. #449 should likely be merged before this PR.

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -6,7 +6,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python: ['3.6', '3.7', '3.8', '3.9', '3.10', 'pypy-3.7']
+        python: ['2.7', '3.5', '3.6', '3.7', '3.8', '3.9', '3.10', 'pypy-3.7']
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -3,8 +3,8 @@
 Release History
 ---------------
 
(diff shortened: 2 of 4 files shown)


cclauss:
To make this ✅ on the legacy Pythons, please edit `tests/requirements.txt` to make similar changes as `pytest` in that file.  Something like:
```
flake8; python_version < '3.6'
flake8>=4.0.1; python_version >= '3.6'
```

JustAnotherArchivist:
Yeah, I already fixed it on another branch, actually, but obviously based on this commit so the 2.7/3.5 tests run at all, so I figured I'd make a separate PR once this is merged to avoid confusion.

JustAnotherArchivist:
Nevermind, it's cleaner to do both in one, so this is superseded by #451.

PR #451, opened 2021-11-25
Fix tests for Python 2.7 and 3.5

JustAnotherArchivist:
* Synchronise the tox and GitHub Actions configs so the Python 2.7 and 3.5 tests are executed again (see #450 for details)
* Fix test dependencies for Python 2.7 and 3.5

Written on top of the v2.2.0 commit, which is why that appears in here. #449 should probably be merged first, but I don't know how GitHub will handle that exactly.


cclauss (approved):
LGTM

PR #456, opened 2021-12-12
Fix possibly faulty values for domain and path in parsed cookies.

### LINKED ISSUE #452: Session creation error with v2.2.0
I am seeing an error with 2.2.0, on Linux, Python3.8, with both the CLI and python library. I have a `~/.config/ia.ini` configuration file with both "s3" and "cookies" auth configured.

Library issue (in pytest output) looks like:

```
fatcat_scholar/work_pipeline.py:91: in __init__
    self.ia_client = internetarchive.get_session()
.venv/lib/python3.8/site-packages/internetarchive/api.py:75: in get_session
    return session.ArchiveSession(config, config_file, debug, http_adapter_kwargs)
.venv/lib/python3.8/site-packages/internetarchive/session.py:114: in __init__
    cookie = create_cookie(ck, cookie_dict[ck],
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

name = 'logged-in-user', value = 'bnewbold%40archive.org', kwargs = {'domain': None, 'path': None}
result = {'comment': None, 'comment_url': None, 'discard': True, 'domain': None, ...}, badargs = set()

    def create_cookie(name, value, **kwargs):
        """Make a cookie from underspecified parameters.
    
        By default, the pair of `name` and `value` will be set for the domain ''
        and sent on every request (this is sometimes called a "supercookie").
        """
        result = {
            'version': 0,
            'name': name,
            'value': value,
            'port': None,
            'domain': '',
            'path': '/',
            'secure': False,
            'expires': None,
            'discard': True,
            'comment': None,
            'comment_url': None,
            'rest': {'HttpOnly': None},
            'rfc2109': False,
        }
    
        badargs = set(kwargs) - set(result)
        if badargs:
            err = 'create_cookie() got unexpected keyword arguments: %s'
            raise TypeError(err % list(badargs))
    
        result.update(kwargs)
        result['port_specified'] = bool(result['port'])
        result['domain_specified'] = bool(result['domain'])
>       result['domain_initial_dot'] = result['domain'].startswith('.')
E       AttributeError: 'NoneType' object has no attribute 'startswith'
```

and the CLI stack trace was:

```
$ ia search 'collection:periodicals collection:sim_microfilm mediatype:texts !noindex:true (pub_type:"Scholarly Journals" OR pub_type:"Historical Journals" OR pub_type:"Law Journals")' -n
  File "/fast/fatcat-scholar/.venv/bin/ia", line 8, in <module>
    sys.exit(main())
  File "/fast/fatcat-scholar/.venv/lib/python3.8/site-packages/internetarchive/cli/ia.py", line 164, in main
    session = get_session(config_file=args['--config-file'],
  File "/fast/fatcat-scholar/.venv/lib/python3.8/site-packages/internetarchive/api.py", line 75, in get_session
    return session.ArchiveSession(config, config_file, debug, http_adapter_kwargs)
  File "/fast/fatcat-scholar/.venv/lib/python3.8/site-packages/internetarchive/session.py", line 114, in __init__
    cookie = create_cookie(ck, cookie_dict[ck],
  File "/fast/fatcat-scholar/.venv/lib/python3.8/site-packages/requests/cookies.py", line 471, in create_cookie
    result['domain_initial_dot'] = result['domain'].startswith('.')
AttributeError: 'NoneType' object has no attribute 'startswith'
```

I speculate that the problem might be passing `domain=cookie_dict.get('domain')` (which could be `None`), and then later checking `result['domain_initial_dot'] = result['domain'].startswith('.')`. One fix would be to use `.get('domain', '')` to have a default value of empty string instead of None. Or to do an `isinstance` check before calling `startswith()`?

It seems suspicious that I would be the first to encounter this, so might be a problem or quirk with my setup.

### LINKED ISSUE #455: Cookie creation with "None" as the domain parameter
https://github.com/jjjake/internetarchive/blob/b347f3f45f2790575cfc8db2657365989ea493b6/internetarchive/session.py#L115

This seems like we need the `domain` parameter to default to `''` rather than `None`. Seems unclear whether this is a [bug in their code](https://github.com/psf/requests/blob/39d0fdd9096f7dceccbc8f82e1eda7dd64717a8e/requests/cookies.py#L468) or ours, but we can account for it easily. 

Side note: the `path` parameter can be `None` with no issue.

maxz:
Using .archive.org as the default domain would be better,
but I'm not sure whether that covers all use cases.
@jjjake would have to chime in here.

Closes #452
Closes #455

### DIFF
--- a/internetarchive/session.py
+++ b/internetarchive/session.py
@@ -112,8 +112,9 @@ class ArchiveSession(requests.sessions.Session):
             if not cookie_dict.get(ck):
                 continue
             cookie = create_cookie(ck, cookie_dict[ck],
-                                   domain=cookie_dict.get('domain'),
-                                   path=cookie_dict.get('path'))
+                                   # .archive.org might be a better domain.
+                                   domain=cookie_dict.get('domain', ''),
+                                   path=cookie_dict.get('path', '/'))
             self.cookies.set_cookie(cookie)
 
         self.secure = self.config.get('general', {}).get('secure', True)
--- a/internetarchive/session.py
+++ b/internetarchive/session.py
@@ -112,8 +112,7 @@ class ArchiveSession(requests.sessions.Session):
             if not cookie_dict.get(ck):
                 continue
             cookie = create_cookie(ck, cookie_dict[ck],
-                                   # .archive.org might be a better domain.
-                                   domain=cookie_dict.get('domain', ''),
+                                   domain=cookie_dict.get('domain', '.archive.org'),
                                    path=cookie_dict.get('path', '/'))
             self.cookies.set_cookie(cookie)
 
--- a/internetarchive/utils.py
+++ b/internetarchive/utils.py
@@ -412,7 +412,7 @@ def parse_dict_cookies(value):
         name, value = item.split('=', 1)
         result[name] = value
     if 'domain' not in result:
-        result['domain'] = ''  # .archive.org might be better.
+        result['domain'] = '.archive.org'
     if 'path' not in result:
         result['path'] = '/'
     return result
--- a/internetarchive/utils.py
+++ b/internetarchive/utils.py
@@ -411,4 +411,8 @@ def parse_dict_cookies(value):
             continue
         name, value = item.split('=', 1)
         result[name] = value
+    if 'domain' not in result:
+        result['domain'] = ''  # .archive.org might be better.
+    if 'path' not in result:
+        result['path'] = '/'
     return result


jjjake:
Thanks once again, @maxz, I appreciate it.

Yeah, I agree that `.archive.org` would be a better default domain. I'll merge once that change is made (happy to merge, and make the change myself if you prefer though). Thanks again!

PR #459, opened 2021-12-17
Make the IA config file path configurable through the IA_CONFIG_FILE environment variable

JustAnotherArchivist:
This change allows setting the IA config file path through an environment variable `IA_CONFIG_FILE` rather than only through the `--config-file` option. When both are present, the option takes precedence.

The motivation for this PR stems from script usage of `ia` and the occasional need for specifying a non-default configuration file path. In the past, I worked around this with a wrapper script which invokes `ia` with `--config-file`, but that approach has its own downsides (e.g. requiring either a hardcoded full path or fragile `PATH` modifications). Another way would be to override `HOME` or `XDG_CONFIG_HOME`, which is all sorts of wrong. I feel like this is a much cleaner solution.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -3,6 +3,13 @@
 Release History
 ---------------
 
+Unreleased
+++++++++++
+
+**Features and Improvements**
+
+- Added support for ``IA_CONFIG_FILE`` environment variable to specify the configuration file path.
+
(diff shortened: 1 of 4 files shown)


jjjake:
Looks good, thank you! Thanks for updating `HISTORY.rst` and adding tests as well! :)

PR #464, opened 2022-01-14
Update README.rst

vvasuki:


jjjake:
I'm OK with adding this documentation to a new section in `docs/source/installation.rst` (and noting exactly what it does and why you might want to do this), but I would prefer not to add it to the top-level README. `ia` is used by a lot of people who might not understand what this does, and should probably stick to the stable releases.

I appreciate the contribution!

jjjake:
Closing, but feel free to reopen if you'd like to add this to `docs/source/installation.rst`. Thanks!

PR #467, opened 2022-01-28
Drop support for Python 2 and 3.5 and remove the corresponding cruft

### LINKED ISSUE #435: Consider dropping Python 2 support

JustAnotherArchivist:
In addition to the obvious changes, this also includes the following noteworthy ones:

- `ia list` had a bug on Python 3, introduced by f52ff26c, where it behaved differently for list-like file metadata on Py2 than Py3. This ports the Py2 behaviour to Py3.
- `internetarchive.utils.map2x` is gone. This was unused since 4f05bebf, only needed for Py2 compatibility, and never part of the public API.
- The `pex-requirements.txt` file didn't match the dependencies in `setup.py` anymore; synchronised those again.

Closes #435


JustAnotherArchivist:
A few possible further steps:

1. Remove `# -*- coding: utf-8 -*-` lines. Source code files default to UTF-8 under Python 3. However, these lines are also used by some editors (but what editor wouldn't default to UTF-8 nowadays‽), and I know some people feel very strongly about them, so I didn't touch them for now.
2. Replace `%` and `.format` string formatting with f-strings (where reasonable given the complexity of the replacements). This is somewhat of a style choice, but it can make code much more readable in my opinion.
3. Replace `dict()` constructor calls with `{}` literals, and likewise for `list` and `set`. I'm not sure if there is a valid reason for the constructors to be used. I'm aware of things like `{'foo': 'bar', **other}`, which I think wasn't possible until PEP 448 and Python 3.5, but I'm referring to plain `dict(key='value')` or `dict([('key', 'value')])`.

Also, regarding the `ia list` change mentioned in the commit message: I've never actually seen this in the wild, so I couldn't test it. An example would be helpful.

@jjjake and others, any thoughts?

And by the way, no specific reason for dropping 3.5 at the same time, but it was already not included in the classifiers and is EOL since September 2020 with no good reason not to upgrade to 3.6 or higher (whereas Py2 had *some* argument due to the incompatibility), so I figured why not clean that up at the same time.

internetarchive/__init__.py
```diff
@@ -79,13 +77,6 @@
...
-    from logging import NullHandler
-except ImportError:
-    class NullHandler(logging.Handler):
-        def emit(self, record):
-            pass
-
 
 log = logging.getLogger(__name__)
-log.addHandler(NullHandler())
+log.addHandler(logging.NullHandler())
```
maxz, line 82: In my eyes `from logging import NullHandler` and then using that name instead of the fully qualified name would be better.

internetarchive/cli/ia_copy.py
```diff
@@ -36,12 +36,11 @@
                                    will not be saved to history/files/$key.~N~
                                    [default: True].
 """
-from __future__ import print_function, absolute_import
 import sys
+import urllib.parse
```
maxz, line 40: I'd rather use `from urllib.parse import quote` and adjust the usages of `urllib.parse.quote` to `quote`.

internetarchive/files.py
```diff
@@ -24,15 +24,11 @@
...
-from __future__ import absolute_import, unicode_literals, print_function
-
 import os
 import sys
 import logging
 import socket
-
-import six.moves.urllib as urllib
-import six
+import urllib.parse
```
maxz, line 31: Same as the above `urllib.parse` comment.

internetarchive/iarequest.py
```diff
@@ -24,21 +24,17 @@
...
 try:
     import ujson as json
 except ImportError:
     import json
 import re
 import copy
 import logging
-
-from six.moves import urllib
+import urllib.parse
```
maxz, line 34: Same as the above `urllib.parse` comment.

internetarchive/item.py
```diff
@@ -34,15 +32,11 @@
...
-    from functools import total_ordering
-except ImportError:
-    from total_ordering import total_ordering
+from functools import total_ordering
 import json
 from copy import deepcopy
 
-from six import string_types
-from six.moves import urllib
+import urllib.parse
```
maxz, line 39: Same as the above `urllib.parse` comment.

maxz (commented):
Just some minor comments regarding cleaner imports.

maxz:
> Remove `# -*- coding: utf-8 -*-` lines.

I would definitely keep the coding line. There are still tons of wrongly configured editors and some of them at least honour that line.

> Replace `%` and `.format` string formatting with f-strings [...] but it can make code much more readable in my opinion.

It tends to make it clunkier in my opinion and does not add noteworthy clarity in any somewhat short case. (Let's say 1-3 lines.)
Above that it can make things clearer, but only if proper names are used in the format string.

> Replace `dict()` constructor calls with `{}` literals, and likewise for `list` and `set`.

I would probably like to keep the constructor calls in the cases where there are no explicit arguments.
Dictionaries and sets use the same kind of brackets for their literals which could confuse people.
As soon as there is data involved, I would also prefer to use the literals instead of the constructors, when the structure is created at compile time.

> And by the way, no specific reason for dropping 3.5 at the same time

It is a good choice, since the libraries that cause us trouble with Python 2.7 also drop their Python 3.5 support at the same time.

JustAnotherArchivist:
Thanks for the review.

> cleaner imports

I much prefer full names over `from` imports. In my own code, I almost never use `from` at all. In my opinion, it's much clearer what's going on when you see `urllib.parse.quote` compared to just `quote`, which could also be a custom function or imported from just about anywhere.

> wrongly configured editors

I'm not a fan of keeping noise in the code just so some people don't have to fix their broken editor configuration...

> [f-string formatting] tends to make it clunkier

I disagree. `f'{foo} is {bar}'` is much more readable to me than `'{} is {}'.format(foo, bar)`.

> I would probably like to keep the constructor calls in the cases where there are no explicit arguments.

Not entirely sure what you mean by this. Are you talking about `list(variable)` where the variable is an iterable? Yeah, certainly, a list comprehension makes no sense for that. I'm talking about statements like `params = dict(rate_limits=1, cmd=cmd)` or `set(['nasa_meta.xml'])` as well as `list()` and `dict()` to create empty containers.

> Dictionaries and sets use the same kind of brackets for their literals which could confuse people.

It can be confusing if you never saw it before, yeah. But then again, so can many other things in Python that share syntax, like iterating over a list vs a generator (where the 'slow' iteration might confuse someone new to the concept), duck typing, or inheritance. I'm not sure that's a valid reason to not use the more common syntax.

By the way, `dict()` is also a fair bit slower than `{}` because it needs to look up the name `dict` in locals, globals, and built-ins first (since it could've been reassigned). Although that's only relevant when you create millions of them, I guess.

jjjake:
Thanks to both of you for jumping on this!

I agree with Maxz on the coding line, please keep it for now.

Regarding the style choices, I'll list my preferences, but I am flexible. I care more about consistency than anything.

- I like from imports. It makes it easy to see what part of a library is used, at the top of the file. It also helps keep the code less crufty, IMO (e.g. long lines with stuff like `'{0} {1}'.format(urllib.parse.quote(x), urllib.parse.quote(y))`). `from foo import bar as foo_bar` can be used when more clarity is needed, too. I can understand both perspectives though.
- I like f-string formatting for clarity.
- I've always tended to use constructor calls rather than literals, for clarity. I don't have a strong preference though.

Dropping 3.5 support sounds good to me, let's do it.

I'll take a closer look next week, and let you know if anything else comes to mind. This is a great start though! Thanks again!

maxz:
> I much prefer full names over `from` imports. In my own code, I almost never use `from` at all. In my opinion, it's much clearer

I can understand the sentiment. I just don't share it. I share the argument that jjjake made in his comment.
My personal standpoint is to always use from-imports unless an ambiguity arises in which case you either use the fully qualified name for one or for both of the ambiguous names.

There might be some edge cases where the fully qualified name really makes sense in other cases too.

> I'm not a fan of keeping noise in the code just so some people don't have to fix their broken editor configuration...

Well, it's a reality we have to deal with and I can guarantee that removing it would lead to wrongly encoded files sooner or later which we may or may not catch at the time of merging.

This comment is a bit ironic if read together with the one above. Since a properly configured editor would show you the definition of any name.

> I disagree. `f'{foo} is {bar}'` is much more readable to me than `'{} is {}'.format(foo, bar)`.

I'm with you there. Personally I dislike the `.format()` style because it looks incredibly clunky and I don't see any clarity gains of `'{} is {}'.format(foo, bar)` over `'%s is %s' % (foo, bar)`. I missed that you were talking about the even newer `f'{foo}'` style. That would indeed make it rather clear.

We might as well standardise on `f'{foo}'` for new code, for all I care.

> Not entirely sure what you mean by this. Are you talking about `list(variable)` where the variable is an iterable? Yeah, certainly, a list comprehension makes no sense for that. I'm talking about statements like `params = dict(rate_limits=1, cmd=cmd)` or `set(['nasa_meta.xml'])` as well as `list()` and `dict()` to create empty containers.

Yes, and also similar cases. Where literal data makes sense I would use the literals instead of the constructors, with the exception of empty sets, where it would use `set()`, but as far as I know there is no literal for empty sets anyways.

I feel somewhat strongly about the imports and would really prefer if we used from-imports.
Regarding the other cases I think we share the same preference.

JustAnotherArchivist:
Yeah, we'll have to agree to disagree on some things there. I feel strongly about imports and encoding lines (and also some other things like tab indentation), but it's not my project, so naturally I'll respect Jake's preferences and decisions on that.

---

@jjjake

> I've always tended to use constructor calls rather than literals, for clarity.

From a bit of grepping, there are significantly more occurrences of `[]` than `list()` in the code by a factor of 2-3 and likewise for `{}` vs `dict()` (only compared the empty values since non-empty ones are harder to grep for). The former are definitely more common in general Python code from what I've seen and are universally understood by anyone who would be likely to look at the code, I think. Sets are a rarer sight in the wild (and in my opinion, they're criminally underused with people mistreating lists and dicts instead), but I'm not sure it makes sense to single them out in terms of consistency. Non-empty sets, i.e. the `{'a', 'b'}` syntax that might confuse some people, are anyway only present in tests as far as I can see (i.e. even fewer and likely more experienced people would ever see them). Empty sets can only be produced with `set()` as there is no literal for that.

---

@maxz

> This comment is a bit ironic if read together with the one above. Since a properly configured editor would show you the definition of any name.

Well, depends on what you mean by 'editor'. A full IDE would do that of course. A text editor like nano, pico, emacs, or vi won't, at least without plugins. Naturally, these editors have no issues with UTF-8, and I assume they've defaulted to UTF-8 at least as long as `LANG` has generally defaulted to a UTF-8 encoding on most Linux distributions.

> Personally I dislike the `.format()` style because it looks incredibly clunky and I don't see any clarity gains of `'{} is {}'.format(foo, bar)` over `'%s is %s' % (foo, bar)`.

That's true, there is no clarity gain. The `.format` syntax is more flexible though; for example, it allows reusing arguments, has many more options for number formatting, and can do right alignment. f-strings go beyond that by allowing expressions in the curly brackets, not just variable names, and yeah, they're much clearer. The only downside I've seen so far is that they can't be evaluated lazily/from a variable storing the format string (yet?). So if you read a format string from a config file or CLI argument, `str.format` is still the best option, sadly.

---

Summarising, I'll:

* change the imports as suggested (except for the `logging` one as commented above, which I'll leave alone for now),
* not touch the coding lines,
* replace string formatting with f-strings where reasonable, and
* wait for a decision on the constructors vs literals.

jjjake:
Your summary sounds great, @JustAnotherArchivist.

Regarding constructors vs literals, yes, I should've been more clear... When I initially wrote most of the code I used literals, and switched at some point to using constructors (the goal being clarity). It doesn't sound like @maxz (please correct me if I'm wrong :)) or anyone else has very strong opinions on this, besides a few specific cases, so I'm fine with going with literals.

Thanks again @JustAnotherArchivist and @maxz I appreciate the contributions and feedback!

maxz:
> When I initially wrote most of the code I used literals, and switched at some point to using constructors (the goal being clarity). It doesn't sound like @maxz (please correct me if I'm wrong :)) or anyone else has very strong opinions on this, besides a few specific cases, so I'm fine with going with literals.

You are correct.

JustAnotherArchivist:
All done and ready for review. :-)

internetarchive/__init__.py
maxz: The original import for Python versions >= 2.7 was `from logging import NullHandler`, `import logging` was there as a fallback, so I would keep the former.

internetarchive/cli/ia.py
```diff
@@ -117,16 +114,16 @@ def main():
...
+        '--config-file': Or(None, str),
+        '--host': Or(None, str),
         '<args>': list,
         '<command>': Or(str, lambda _: 'help'),
     })
     try:
         args = s.validate(args)
     except SchemaError as exc:
-        print('{0}\n{1}'.format(str(exc), printable_usage(__doc__)), file=sys.stderr)
+        print(f'{exc!s}\n{printable_usage(__doc__)}', file=sys.stderr)
```
maxz, line 126: The cast to string here is very terse, do we really want this instead of `str(exc)`?

If we switch to the `f''` syntax we might as well embrace it fully, but this is still something that we should consider first.

internetarchive/cli/ia_tasks.py
```diff
@@ -96,28 +94,25 @@ def main(argv, session):
                                 data=data)
         j = r.json()
         if j.get('success'):
-            print('success: {}'.format(j.get('value', dict()).get('log')))
+            print(f'success: {j.get("value", {}).get("log")}')
```
maxz, line 97: This is a case where I feel the new syntax stretches a bit far. While the previous version was not any clearer, it maybe should have been written differently in the first place and we should decide whether we want such "complex" statements within the format strings.

maxz (approved):
I did not check the Makefile in detail.

I looked over everything else and it looks good.
The logging import should probably be changed before merging.
All other comments are more about how we want to deal with similar cases in the future and especially @jjjake should take a look at them.

If we have to deal with long statements including the new format strings, it also gets a bit more complicated to split them over multiple lines.
I would use implicit string merging within parentheses in those cases, the only downside is that every partial string has to be prefixed with an f:
```
(f'This is a {foo} example'
 f'of a {bar} multiline format string.')
```

internetarchive/__init__.py
JustAnotherArchivist: The `import logging` was and is also used by `logging.getLogger`.

internetarchive/cli/ia_tasks.py
JustAnotherArchivist: Yeah, I thought about this one a bit while writing it. The alternative would be to put it in a variable which is then used in the f-string. I'm not really sure that's better though.

internetarchive/cli/ia.py
JustAnotherArchivist: Definitely in favour of embracing it. Another option would be to get rid of these string casts because variables always get cast to string inside an f-string anyway when you don't specify a conversion operator.

jjjake: Good point, @JustAnotherArchivist, let's get rid of these string casts when they're not needed.

When it is needed, I agree that we should embrace `f''` syntax fully.

internetarchive/cli/ia_tasks.py
jjjake: Yeah, this definitely should've been written differently in the first place, IMO. I think using a variable would be preferable, I'd probably do something like this:

```
task_log_url = j.get("value", {}).get("log")
print(f'success: {task_log_url}')
```

The variable name, `task_log_url`, at least gives some idea what is being printed here. No worries if you don't deal with this in this PR. I can fix it after merging.

jjjake:
I responded to the comments above, and will do a full review tomorrow. Thanks again to both of you!

internetarchive/cli/ia.py
JustAnotherArchivist: Minor correction: the default conversion happens using the variable's `__format__` method. Because neither `schema.SchemaError` nor anything in its parent class tree defines such a method, the top-level `object.__format__` is used, which [simply returns `str(self)`](https://github.com/python/cpython/blob/a0e55a571cf01885fd5826266c37abaee307c309/Objects/typeobject.c#L5532). The same thing holds for the other cases where the `!s` conversion is currently used. Will remove shortly.

maxz (approved):
Looks good. Thank you for taking the time to move everything closer to Python 3, @JustAnotherArchivist.

There are still some places in code which I wrote which can be changed to fully embrace Python 3 (I will change those after this PR is merged) and there are probably some similar places elsewhere in the library, but we will catch those over time and this is already a giant leap toward shedding the cruft.

JustAnotherArchivist:
Oh yes, this is definitely by no means complete, just a first major step in the right direction. While going through the code, I also spotted a few things I'd like to clean up later but didn't want to include here since they're unrelated to the Python 2 removal.

jjjake:
All looks great, merged! Thanks again @JustAnotherArchivist and @maxz! I appreciate the feedback and contributions. :)

I'll ship v3.0.0 next week with these changes. Let me know if you think we should make any other changes before that.

PR #470, opened 2022-02-03
Clean up JSON package importing to use ujson throughout when available

JustAnotherArchivist:
* ujson was only being used in some places previously and not others.
* The upload CLI used the undocumented `json.decoder.JSONDecodeError`.
* `tests/test_item.py::test_download_checksum` was broken when run with ujson due to different but equivalent encoding.


JustAnotherArchivist:
The flake8 error comes from bfd8ef7d, not the changes in this commit.

JustAnotherArchivist:
The flake8 error comes from bfd8ef7d, not the changes in this commit.

PR #476, opened 2022-02-06
Clean up usage of `open()`

JustAnotherArchivist:
* Get rid of `io.open`, which is an alias for `open` in Python 3.
* Remove obsolete 'U' mode, which has no effect in Python 3.
* Use context managers so the files get closed properly.


jjjake:
Thanks again @JustAnotherArchivist, I really appreciate all the contributions!

PR #481, opened 2022-02-09
Revert a change which was made to accommodate Python 2.

maxz:
Change back to using `re.fullmatch` instead of an equivalently anchored `re.match`, to be more idiomatic.


jjjake:
Thanks @maxz!

PR #485, opened 2022-02-12
Item status check: error on maximum item size exceedance and test with specific identifier/access key

### LINKED ISSUE #293: upload command's status check flag doesn't check if item is full

JustAnotherArchivist:
Fixes #293

### DIFF
--- a/internetarchive/cli/ia_upload.py
+++ b/internetarchive/cli/ia_upload.py
@@ -80,6 +80,9 @@ from internetarchive.utils import (InvalidIdentifierException, JSONDecodeError, 
                                    is_valid_metadata_key, json, validate_s3_identifier)
 
 
+MAX_ITEM_SIZE = 2**40  # 1 TiB
+
+
 def _upload_files(item, files, upload_kwargs, prev_identifier=None, archive_session=None):
     """Helper function for calling :meth:`Item.upload`"""
     responses = []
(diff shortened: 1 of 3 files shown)


JustAnotherArchivist:
Also replaces the abandoned PR #297

jjjake:
Thanks again @JustAnotherArchivist!

This looks good, but I'm actually going to look into what it'd take to get limit info from s3.us.archive.org rather than hard-coding it here. I'll keep you posted.

internetarchive/cli/ia_upload.py
```diff
@@ -160,19 +163,22 @@ def main(argv, session):
...
+    if args['<identifier>']:
+        item = session.get_item(args['<identifier>'])
     if args['--status-check']:
-        if session.s3_is_overloaded():
+        if session.s3_is_overloaded(identifier=args['<identifier>'], access_key=session.access_key):
             print(f'warning: {args["<identifier>"]} is over limit, and not accepting requests. '
                   'Expect 503 SlowDown errors.',
                   file=sys.stderr)
             sys.exit(1)
+        elif item.item_size >= MAX_ITEM_SIZE:
```
cclauss, line 174:
```suggestion
        elif item.item_size > MAX_ITEM_SIZE:
```

internetarchive/cli/ia_upload.py
```diff
@@ -160,19 +163,22 @@ def main(argv, session):
             sys.exit(1)
 
     # Status check.
+    if args['<identifier>']:
+        item = session.get_item(args['<identifier>'])
```
cclauss, line 167: Do we really want to `get` an item that could be 1TB or more _before_ we do a status-check?

internetarchive/cli/ia_upload.py
JustAnotherArchivist: This would require some testing whether IA's servers still accept any upload (including an empty file) if the item is exactly 1 TiB. Might be tricky though since I think the metadata files, which get modified after every upload, also count towards the item size.

internetarchive/cli/ia_upload.py
JustAnotherArchivist: Well, we need the `Item` object for both the size status check and the actual upload. While this structure means we needlessly fetch the item metadata when S3 is overloaded, it avoids more complicated conditions (e.g. first run the S3 overload check if `--status-check` is present, then fetch item metadata if the identifier is present, then check the item size if both are present, then exit successfully if `--status-check` is present), which in my opinion leads to less readable code. Alternatives are two lines with `get_item` calls, which is just as ugly, or some sort of lazy evaluation, which is somewhat complicated to implement. So I found this to be the least awkward solution.

PR #497, opened 2022-04-10
Search._scrape() verify that num_found is an int

cclauss:
On line 160 we compare `i` and `num_found` so let's ensure that they are both ints.


jjjake:
Thanks, cclauss!

PR #503, opened 2022-04-24
utils.py: `type: ignore` a few mypy issues

cclauss:
Fixed in https://github.com/ultrajson/ultrajson/pull/498


JustAnotherArchivist:
This requires a bump of the ujson version.

PR #504, opened 2022-04-24
Fix a few mypy issues in docs/source/conf.py

cclauss:


docs/source/conf.py
```diff
@@ -12,8 +12,10 @@
 # All configuration values have a default; values that are commented out
 # serve to show the default.
 
-import sys
+from __future__ import annotations
```
maxz, line 15: As far as I know this import is not required in 3.7 and only enables ["postponed evaluation"](https://docs.python.org/3/whatsnew/3.7.html#pep-563-postponed-evaluation-of-annotations) which is a feature of later versions.

The annotations you added do not make use of the postponed evaluation.

docs/source/conf.py
```diff
@@ -79,7 +81,7 @@
 
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
-exclude_patterns = []
+exclude_patterns: str = []
```
maxz, line 84: Should this not rather be `list[str]` instead of `str`?

maxz (changes requested):

docs/source/conf.py
cclauss: Why resist importing from `__future__`...  It is part of the language.

$ `docker run --rm -it python:3.7 bash ` # Wait for that container to open...
# `apt-get -qq update && apt-get install -y python3 && python3 --version ` # Python 3.7.13...
# `python3.7 -c "d: dict | None = None" ` # TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'

docs/source/conf.py
cclauss: Yes.  Thanks.  Fixing.

maxz (approved):

maxz:
The new title of the pull request does not properly reflect the changes. I think you accidentally copied the wrong explanation over while changing the title of your other PR.

PR #506, opened 2022-04-25
isort your imports, so you don't have to

cclauss:
# [`isort --profile black .`](https://github.com/PyCQA/isort#readme)


maxz:
Great change. I was working on enforcing most checks in `lint_python.yml` some months ago, but dropped that work after I wanted to make it easy for every contributor to test and enforce the tests before committing and went down the PEP 518 rabbit hole, trying to get everything into one of the new dependency management systems (Poetry in this particular case). It just became a too much work at once.

I always wanted to ask whether you could approach that problem since you seem to have particular expertise in exactly that field.

PR #516, opened 2022-04-29
Add Python type hints to utils.py

cclauss:
@JustAnotherArchivist @maxz Your reviews, please.

If you want, grab a file and dig in...

### DIFF
--- a/internetarchive/utils.py
+++ b/internetarchive/utils.py
@@ -25,11 +25,14 @@ This module provides utility functions for the internetarchive library.
 :copyright: (C) 2012-2022 by Internet Archive.
 :license: AGPL 3, see LICENSE for more details.
 """
+from __future__ import annotations
+
 import hashlib
 import os
 import re
 import sys
 from collections.abc import Mapping
+from typing import Iterable
 from xml.dom.minidom import parseString
 
 # Make preferred JSON package available via `from internetarchive.utils import json`
@@ -43,7 +46,7 @@ except ImportError:
     JSONDecodeError = json.JSONDecodeError  # type: ignore
 
 
-def deep_update(d, u):
+def deep_update(d: dict, u: Mapping) -> dict:
     for k, v in u.items():
         if isinstance(v, Mapping):
             r = deep_update(d.get(k, {}), v)
@@ -57,7 +60,7 @@ class InvalidIdentifierException(Exception):
     pass
 
 
-def validate_s3_identifier(string):
+def validate_s3_identifier(string: str) -> bool:
     legal_chars = 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789._-'
     # periods, underscores, and dashes are legal, but may not be the first
     # character!
@@ -78,7 +81,7 @@ def validate_s3_identifier(string):
     return True
 
 
-def needs_quote(s):
+def needs_quote(s: str) -> bool:
     try:
         s.encode('ascii')
     except (UnicodeDecodeError, UnicodeEncodeError):
@@ -86,16 +89,16 @@ def needs_quote(s):
     return re.search(r'\s', s) is not None
 
 
-def norm_filepath(fp):
+def norm_filepath(fp: bytes | str) -> str:
+    if isinstance(fp, bytes):
+        fp = fp.decode('utf-8')
     fp = fp.replace(os.path.sep, '/')
     if not fp.startswith('/'):
         fp = f'/{fp}'
-    if isinstance(fp, bytes):
-        fp = fp.decode('utf-8')
     return fp
 
 
-def get_md5(file_object):
+def get_md5(file_object) -> str:
     m = hashlib.md5()
     while True:
         data = file_object.read(8192)
@@ -106,7 +109,7 @@ def get_md5(file_object):
     return m.hexdigest()
 
 
-def chunk_generator(fp, chunk_size):
+def chunk_generator(fp, chunk_size: int):
     while True:
         chunk = fp.read(chunk_size)
         if not chunk:
@@ -114,7 +117,7 @@ def chunk_generator(fp, chunk_size):
         yield chunk
 
 
-def suppress_keyboard_interrupt_message():
+def suppress_keyboard_interrupt_message() -> None:
     """Register a new excepthook to suppress KeyboardInterrupt
     exception messages, and exit with status code 130.
 
@@ -131,14 +134,14 @@ def suppress_keyboard_interrupt_message():
 
 
 class IterableToFileAdapter:
-    def __init__(self, iterable, size):
+    def __init__(self, iterable, size: int):
         self.iterator = iter(iterable)
         self.length = size
 
-    def read(self, size=-1):  # TBD: add buffer for `len(data) > size` case
+    def read(self, size: int = -1):  # TBD: add buffer for `len(data) > size` case
         return next(self.iterator, b'')
 
-    def __len__(self):
+    def __len__(self) -> int:
         return self.length
 
 
@@ -153,7 +156,7 @@ class IdentifierListAsItems:
         self._items = [None] * len(self.ids)
         self.session = session
 
-    def __len__(self):
+    def __len__(self) -> int:
         return len(self.ids)
 
     def __getitem__(self, idx):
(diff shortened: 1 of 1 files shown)


JustAnotherArchivist (approved):
lgtm

internetarchive/utils.py
```diff
@@ -43,7 +45,7 @@
     JSONDecodeError = json.JSONDecodeError  # type: ignore
 
 
-def deep_update(d, u):
+def deep_update(d: dict, u: Mapping) -> dict:
```
maxz, line 48: The recommended way seems to be to use the more abstract type for the arguments and the narrower types for the return values. So `-> dict` would be as recommended but if we follow the Python documentation's recommendation, then `d` should also have the type `Mapping`.

The same is true for all other functions in this file.

internetarchive/utils.py
```diff
@@ -78,24 +80,24 @@ def validate_s3_identifier(string):
...
+def needs_quote(s: str) -> bool:
     try:
         s.encode('ascii')
     except (UnicodeDecodeError, UnicodeEncodeError):
         return True
     return re.search(r'\s', s) is not None
 
 
-def norm_filepath(fp):
+def norm_filepath(fp: bytes | str) -> str:
```
maxz, line 91: Not for you to fix here, just as a general remark: We should change this due to having dropped the Python 2 support.

internetarchive/utils.py
```diff
@@ -196,7 +198,7 @@ def _get_tag_text(tag_name, xml_obj):
         return str(xml_str)
 
 
-def get_file_size(file_obj):
+def get_file_size(file_obj) -> int | None:
```
maxz, line 201: Maybe Optional[int] should be preferred as the return type? I'm not quite sure about which would be more readable. I guess `int | None` would be more readable to people who had no contact with the type annotations before while `Optional[int]` would be clearer to people who are more used to the type annotations.

internetarchive/utils.py
```diff
@@ -332,7 +334,7 @@ def remove_none(obj):
         return obj
 
 
-def delete_items_from_dict(d, to_delete):
+def delete_items_from_dict(d: dict, to_delete):
```
maxz, line 337: A complicated case. For now this seems fitting but I think the whole function should be refactored.

`Dict | List` seems to be a fitting return type for now.
`d` would also have to be `Dict | List`, `Mapping | Iterable` would be preferred, but since the types are checked explicitly in this case, I would stick with `Dict | List` for now.

internetarchive/utils.py
```diff
@@ -363,7 +365,7 @@ def is_valid_metadata_key(name):
     return bool(re.fullmatch(r'[A-Za-z][.\-0-9A-Za-z_]+(?:\[[0-9]+\])?', name))
 
 
-def merge_dictionaries(dict0, dict1, keys_to_drop=None):
+def merge_dictionaries(dict0: dict, dict1: dict, keys_to_drop=None) -> dict:
```
maxz, line 368: `dict0: Mapping, dict1: Mapping, keys_to_drop: Optional[Iterable] = None`

maxz (changes requested):

internetarchive/utils.py
JustAnotherArchivist: That sounds very odd to me. The return type here is always `type(d)` since `d` is simply modified in-place and returned...

We could just get rid of the `return` entirely and document it to be an in-place update. The function is only used in one place (`internetarchive.config.get_config`), and the return value is discarded there.

internetarchive/utils.py
JustAnotherArchivist: I do think that `int | None` is more explicit and readable, especially now that `typing.Required` and `typing.NotRequired` is also a thing, albeit in other contexts ([PEP 655](https://peps.python.org/pep-0655/)).

This is [PEP 604](https://peps.python.org/pep-0604/), by the way, and I'm not entirely sure how it behaves on Python 3.7 to 3.9 (though it clearly isn't an error with the `__future__` import).

maxz: Well, `int | None` is certainly more explicit or obvious.

Although I think that `Optional[int]` communicates the intent better and also fits better with the practices in type theory and type systems of other languages like ML, Haskell or C++.
So I'm undecided on this.

`None` is essentially often just a value used as input or output if something is optional and it does not really matter that it is `None`, but rather that it is optional.

internetarchive/utils.py
maxz: I was merely quoting the [relevant part of the documentation](https://docs.python.org/3/library/typing.html#typing.Dict) and it makes a lot of sense to me combined with the duck typing approach of Python, since otherwise you would lose every advantage of duck typing.

With this particular function we have a very special case due to your described in-place mutation, where it would also make sense to keep `dict` as type. Probably it would be best to document the in-place mutation, return nothing, type the function as `-> None` and to use `Mapping` as input type, since that will ensure the appropriate interface for the mutation and the caller will have the exact type information.

cclauss: Modifying a function the returns a value to be a function that returns no value is a breaking change (semver major) so I would prefer to avoid that.

internetarchive/utils.py
cclauss: `pyupgrade --py310-plus` will remove the `annotations` import  and then drop both `typing.Optional` and `typing.Union` so let’s go that way too.

internetarchive/utils.py
maxz: It seems very unlikely that someone external is using this function. But you are correct.

Therefore keeping the signature which you already used would be the best course for now.

internetarchive/utils.py
maxz: Interesting. I did not know that Optional would be deprecated or at least discouraged.
But lets definitely go with `|` then.

internetarchive/utils.py
cclauss: `pyupgrade --py310-plus` will remove the `annotations` import  and then lowercase both `dict` and `list` so let’s go that way too.

The function name pushes me too keep the argument type as `dict`.  Can items be deleted from a Mapping or would we need to specify as MutableMappling?

JustAnotherArchivist: Yeah, it'd have to be `MutableMapping` for the `__delitem__` method.

maxz: It would have to be a `MutableMapping`.

But I would use `dict | list` for now, since there are those explicit type checks anyways.

maxz: Especially since you also cannot delete items from iterables. I'm not sure what the appropriate base class for those would be. There probably is none since the iterable interface is rather basic and it would not make much sense to have a common mutation method.

cclauss, line 337:
```suggestion
def delete_items_from_dict(d: dict | list, to_delete):
```

internetarchive/utils.py
cclauss: https://github.com/asottile/pyupgrade#pep-604-typing-rewrites

internetarchive/utils.py
cclauss: Mypy: internetarchive/utils.py:397: error: "Mapping[Any, Any]" has no attribute "copy"

cclauss:
@maxz Are you able to approve this PR in its current form so that @jjjake can review it?

internetarchive/utils.py
```diff
@@ -363,7 +370,11 @@ def is_valid_metadata_key(name):
     return bool(re.fullmatch(r'[A-Za-z][.\-0-9A-Za-z_]+(?:\[[0-9]+\])?', name))
 
 
-def merge_dictionaries(dict0, dict1, keys_to_drop=None):
+def merge_dictionaries(
+    dict0: dict,
+    dict1: dict,
```
cclauss, line 375:
```suggestion
    dict1: Mapping,
```

maxz (approved):

internetarchive/utils.py
maxz: Some of those interfaces are defined ratherstrangely ...

cclauss:
The test failure seems to be a fluke.

internetarchive/utils.py
JustAnotherArchivist: Regarding the more general issue of modifying interfaces, I'm not sure the `utils` module is even considered part of the public API. It isn't mentioned in the docs, for example. Similarly for the CLI, which I can't imagine being intended to be imported from elsewhere. It may be worth moving them to `internetarchive._utils` and `._cli` instead if they should be treated as internal API. (And yeah, such a change should probably happen on a major version increment.)

cclauss: The recommendation in gradual typing is to start by typing the innermost code first and end its the outermost code last.  I do not know this codebase well enough to know which is which but typing is for both internal and external communications.  I started with a file that I was having trouble with on another project.  Next is this file which is relied on by multiple internal files.  What should come next…

PR #517, opened 2022-04-30
Add Python type hints to auth.py and config.py

cclauss:
@JustAnotherArchivist, @maxz Your reviews, please.

### DIFF
--- a/internetarchive/config.py
+++ b/internetarchive/config.py
@@ -23,10 +23,12 @@ internetarchive.config
 :copyright: (C) 2012-2019 by Internet Archive.
 :license: AGPL 3, see LICENSE for more details.
 """
-import configparser
-import errno
+from __future__ import annotations
+
 import os
 from collections import defaultdict
+from configparser import RawConfigParser
+from typing import Mapping
 
 import requests
 
@@ -35,8 +37,7 @@ from internetarchive.exceptions import AuthenticationError
 from internetarchive.utils import deep_update
 
 
-def get_auth_config(email, password, host=None):
-    host = host if host else 'archive.org'
+def get_auth_config(email: str, password: str, host: str = 'archive.org') -> dict:
     u = f'https://{host}/services/xauthn/'
     p = {'op': 'login'}
     d = {'email': email, 'password': password}
@@ -70,7 +71,7 @@ def get_auth_config(email, password, host=None):
     return auth_config
 
 
-def write_config_file(auth_config, config_file=None):
+def write_config_file(auth_config: Mapping, config_file=None):
     config_file, is_xdg, config = parse_config_file(config_file)
 
     # S3 Keys.
@@ -107,7 +108,7 @@ def write_config_file(auth_config, config_file=None):
 
 
 def parse_config_file(config_file=None):
-    config = configparser.RawConfigParser()
+    config = RawConfigParser()
 
     is_xdg = False
     if not config_file:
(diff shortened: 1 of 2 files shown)


maxz:
I will review it next weekend if it has not been reviewed and merged by then. I'm somewhat busy right now.

internetarchive/config.py
```diff
@@ -70,7 +71,7 @@ def get_auth_config(email, password, host=None):
     return auth_config
 
 
-def write_config_file(auth_config, config_file=None):
+def write_config_file(auth_config: Mapping, config_file=None):
```
maxz, line 74: You might want to use `mapping` instead of `Mapping`, depending on how you wrote it in the other PRs, but if I recall correctly you decided to use the lowercase version of all the types.

`-> string`, but should probably be some proper path type in the future, so maybe keep it the way it is for now.

internetarchive/config.py
```diff
@@ -157,14 +158,14 @@ def parse_config_file(config_file=None):
     return (config_file, is_xdg, config)
 
 
-def get_config(config=None, config_file=None):
-    _config = {} if not config else config
+def get_config(config=None, config_file=None) -> dict:
```
maxz, line 161: `config: mapping | None = None`. For `config_file`the same as for the comment above applies. It would currently be `string | None`, but we most likely want it to be a path in the future so we can skip that one for now.

maxz (approved):
I think the the type of the `config` parameter in `get_config` should be adjusted, the rest looks good.

internetarchive/config.py
cclauss: ### Try `Mapping | None` ...
internetarchive/config.py:166: error: Incompatible return value type (got "Mapping[Any, Any]", expected "Dict[Any, Any]")
internetarchive/config.py:169: error: Item "Mapping[Any, Any]" of "Optional[Mapping[Any, Any]]" has no attribute "sections"
internetarchive/config.py:169: error: Item "None" of "Optional[Mapping[Any, Any]]" has no attribute "sections"
internetarchive/config.py:171: error: Item "None" of "Optional[Mapping[Any, Any]]" has no attribute "items"
internetarchive/config.py:171: error: Too many arguments for "items" of "Mapping"
Found 5 errors in 1 file (checked 1 source file)

---
### Try `RawConfigParse | None` ...
internetarchive/config.py:162: error: Need type annotation for "_config"
internetarchive/config.py:166: error: Incompatible return value type (got "Union[RawConfigParser, Dict[Any, Any]]", expected "Dict[Any, Any]")
internetarchive/config.py:169: error: Item "None" of "Optional[RawConfigParser]" has no attribute "sections"
internetarchive/config.py:171: error: Item "None" of "Optional[RawConfigParser]" has no attribute "items"
internetarchive/session.py:101: error: Argument 1 to "get_config" has incompatible type "Optional[Dict[Any, Any]]"; expected "Optional[RawConfigParser]"
Found 5 errors in 2 files (checked 1 source file)

### So I left is unspecified.

cclauss:
Thanks for the review @maxz but given the comments above, I do not see how to improve these changes.

There are other PRs to review if you get some bandwidth: https://github.com/jjjake/internetarchive/pulls/cclauss

maxz:
I see. It looked like a `Mapping` would cover it, but apparently not.
I don't fully grok all intricacies of the type annotations yet.

You also seem to be getting better with them with each file. While this PR did not return any custom classes, your other PRs do in places.

I already started taking a look at the other PRs but they contain so many changes that I wanted to give them a second viewing. I did not see anything special when going through the changes yet and really liked that you properly used the default arguments to get rid of some of the cruft. And then I wanted to properly look up some points like the one where you were not sure whether that one variable would better be a boolean.

PR #521, opened 2022-05-07
Add Python type hints to internetarchive/cli/ia_metadata.py

cclauss:
```
mypy.....................................................................Failed
internetarchive/cli/ia_metadata.py:138: error: Item "str" of "Union[List[Any], str]" has no attribute "append"
internetarchive/cli/ia_metadata.py:141: error: Item "str" of "Union[List[Any], str]" has no attribute "append"
internetarchive/cli/ia_metadata.py:196: error: Argument 1 to "append" of "list" has incompatible type "bool"; expected "Response"
internetarchive/cli/ia_metadata.py:199: error: Argument 1 to "append" of "list" has incompatible type "bool"; expected "Response"
Found 4 errors in 1 file (checked 1 source file)
```

### DIFF
--- a/internetarchive/cli/ia_metadata.py
+++ b/internetarchive/cli/ia_metadata.py
@@ -54,23 +54,21 @@ import os
 import sys
 from collections import defaultdict
 from copy import copy
-from typing import Mapping, cast
-from urllib.request import Request
+from typing import Mapping
 
 from docopt import docopt, printable_usage  # type: ignore
 from requests import Response
 from schema import And, Or, Schema, SchemaError, Use
 
+from internetarchive import item, session
 from internetarchive.cli.argparser import (get_args_dict,
                                            get_args_dict_many_write,
                                            get_args_header_dict)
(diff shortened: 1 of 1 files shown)


internetarchive/cli/ia_metadata.py
```diff
@@ -226,15 +232,16 @@ def main(argv, session):
             else:
                 responses.append(modify_metadata(item, metadata, args))
             if (i + 1) == len(args['<identifier>']):
-                if all(r.status_code == 200 for r in responses):
+                if all(cast(Response, r).status_code == 200 for r in responses):
```
maxz, line 235: I'm not too fond of this cast, since it is really strange for the readability. But I guess `mypy` strictly demands it.

maxz (approved):
Looks fine.

internetarchive/cli/ia_metadata.py
cclauss: You are right.  I removed it in favor of a `# type: ignore` which is easier on the reader.

PR #524, opened 2022-05-11
Update docs based on Python type hints

cclauss:
It was quite easy to add https://pypi.org/project/sphinx-autodoc-typehints to `docs/requirements.txt` and then generate docs based on Python type hints that look like this:

Screenshot 2022-05-11 at 07 30 29

---

This enables us to remove a fair amount of repetitive (and somewhat erroneous) typing information from method docstrings.  These lines can all be remove from docstrings: `grep ":type " **/*.py` and `grep ":rtype:" **/*.py`

@JustAnotherArchivist @maxz ... Might be of interest.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,3 +1,3 @@
-alabaster==0.7.6
-docutils==0.12
-Sphinx==1.2.2
+alabaster==0.7.12
+docutils==0.18.1
+Sphinx==4.5.0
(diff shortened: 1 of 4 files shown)


cclauss:
1. Change all occurrences of `:rtype:` to `:returns:`
2. Delete lines that contain "`:type `"  (only if the Python type hint is present).
3. Change all occurrences of `: (optional) ` with `: `
4. Go back thru the `:returns:` and clean up any duplicate entries.
```

```

maxz (approved):
Nice. I never particularly liked to also have to specify type information in the docstrings.

PR #526, opened 2022-05-21
Add Python type hints to several more files

cclauss:
@maxz Your review please if you have some bandwidth.


maxz:
Jake merges them faster than I can review them. Haha.

I reviewed about half of it so far and most of it looked good. I still wanted to look deeper into some of the `or ""` places or similar and think a bit about them since I'm not quite sure whether that is really the best default value there.
But I was also not sure whether you are following some bigger plan to slim the passed parameters down or to get rid of some normalization in the call chain, so that not every function has to normalize its arguments separately.

PR #532, opened 2022-05-25
Enable black formatting in pre-commit

cclauss:
Add https://github.com/pre-commit/pre-commit-hooks#double-quote-string-fixer because this repo tends to favor single-quotes instead of black-standard double-quotes.

Remove [`--diff`](https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#diffs) option which prevents `black` from re-formatting our Python files.

Move`isort` configuration into `setup.cfg`.

Move `black` configuration out of `setup.cfg` because `black` only supports config via [`pyproject.toml`](https://github.com/psf/black/blob/main/docs/usage_and_configuration/the_basics.md#configuration-via-a-file).

```
pre-commit autoupdate
pre-commit run --all-files
```


maxz:
It is awesome that we reached this state. I was working on this some months ago when Black reached stable status. Together with sorted imports and so on, minus the type declarations which we have now, but dropped it because it was just such a huge amount of work and getting it all to work together with Poetry was a pain.

I also did not want to do it via pre-commit hooks at the time, but you found a good way to do it with pre-commit hooks.

This should remove virtually all formatting related corrections on new pull requests.

cclauss:
I have never gotten Poetry to sing for me.

eggplants (approved):

PR #527, opened 2022-05-21
Some improvements

eggplants:
- move package information into setup.cfg
- add extra dependencies (`all`, `dev`, `test`, `types`)
- add typed markers (`py.typed`)
- fix mypy errors
- improve workflow files (use cache for pip, names)

PR #533, opened 2022-06-10
GitHub Action to install ia with and without tqdm

cclauss:
Proof that `tqdm` MUST be installed before `internetarchive`

@eggplants Was there any change #527 that could cause pip to not recognize that `tqdm` is a pre-requisite for `internetarchive`?

### DIFF
--- a/.github/workflows/install_internetarchive.yml
+++ b/.github/workflows/install_internetarchive.yml
@@ -1,0 +1,14 @@
+name: install_internetarchive
+on:
+  pull_request:
+  push:
+jobs:
+  install_internetarchive:
+    runs-on: ubuntu-latest
+    steps:
+      - run: pip install internetarchive
+  install_tqdm_and_then_internetarchive:
+    runs-on: ubuntu-latest
+    steps:
+      - run: pip install tqdm  # This line is REQUIRED
+      - run: pip install internetarchive


eggplants:
> Proof that tqdm MUST be installed before internetarchive

What errors did you encounter? Now I try to install `internetarchive` and no error occurred in my env without tqdm.

eggplants:
Oh, I just noticed the error in CI. I'll investigate it.

eggplants:
Please try to add build matrix to CI and test each python3.7-10 with setup-python.
Maybe caused by setuptools or pip?

eggplants:
The problem seems to be that setuptools is reading the version from `__init__.py` dynamically instead of statically.

cclauss:
So what change would you recommend for this repo?

eggplants:
Workaround: move `__version__.py` from `__init__.py` into `_version.py` (or `__version__.py` if you prefer) and change `version`'s attr in `setup.py`.

cclauss:
@jjjake Your thoughts, please.

eggplants:
Another workaround: https://github.com/microsoft/msticpy/commit/55ae50ac5039457c06025a6c119754a61d681eb4

jjjake:
I'm in favor of sticking version in `_version.py` or `__version__.py` (no preference) rather than using `re`.

I also noticed that you need to manually install `requests`, `tqdm`, and `jsonpatch` manually before running `pip install internetarchive` on Ubuntu Focal, otherwise it will fail. Alternatively, `pip install --upgrade setuptools` seems to do the trick as well. Is this related to #527 as well? The setuptools version it was failing on for me was 45.2.0 (Python 3.8.10). Upgrading to 62.4.0 fixed it.

Is there something we can do to still support older versions of setuptools?

eggplants:
Since #534 has been merged, I think this PR can be closed. Sorry for my mistake and thank you for your reports and tests.

PR #537, opened 2022-06-19
upload: Expand a period to the contents of the current directory.

### LINKED ISSUE #220: Special-case uploading current directory
If I try `ia upload SOME-ITEM .` when there is a file with a dot in the current directory, I get:

    error uploading /./RDS_ios.iso.sha: Invalid Argument - when key is split on /, result cannot contain . or ..

Work around is to `ia upload SOME-ITEM *`. As a low-priority feature request, would be great to say `.`.

### LINKED ISSUE #535: `--checksum` does not work when used with `--no-derive`
When i execute `ia upload --checksum --verify --retries 10 --no-backup --no-derive something .` it's just starts uploading right away without checking checksums.

Windows 11, internetarchive 3.0.0 in powershell

maxz:
The previous behavior when specifying a period as a wildcard was faulty.
Now a period is properly expanded to the contents of the current directory.

ia_upload.py seemed to be the most sensible location for this fix, since it
is the closest to receiving the path and so that the behavior of the library
does not change.

Closes #220, fixes #535.

### DIFF
--- a/internetarchive/cli/ia_upload.py
+++ b/internetarchive/cli/ia_upload.py
@@ -237,6 +237,9 @@ def main(argv, session):
             local_file.seek(0)
         else:
             local_file = args['<file>']
+            # Properly expand a period to the contents of the current working directory.
+            if local_file and local_file[0] == '.':
+                local_file = os.listdir('.')
 
         if isinstance(local_file, (list, tuple, set)) and args['--remote-name']:
             local_file = local_file[0]
--- a/internetarchive/cli/ia_upload.py
+++ b/internetarchive/cli/ia_upload.py
@@ -238,8 +238,9 @@ def main(argv, session):
         else:
             local_file = args['<file>']
             # Properly expand a period to the contents of the current working directory.
-            if local_file and local_file[0] == '.':
-                local_file = os.listdir('.')
+            if '.' in local_file:
+                local_file = [p for p in local_file if p != '.']
+                local_file = os.listdir('.') + local_file
 
         if isinstance(local_file, (list, tuple, set)) and args['--remote-name']:
             local_file = local_file[0]
--- a/internetarchive/cli/ia_upload.py
+++ b/internetarchive/cli/ia_upload.py
@@ -114,7 +114,7 @@ def _upload_files(item, files, upload_kwargs, prev_identifier=None, archive_sess
     return responses
 
 
-def main(argv, session):
+def main(argv, session):  # noqa: C901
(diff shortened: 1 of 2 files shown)


JustAnotherArchivist:
This breaks when multiple files/directories are provided. If the dot is the first one, all others would get ignored. And if it isn't the first one, it would not get expanded.

maxz:
Yes, I know. It seemed to be the best compromise here.

I deemed it to be unlikely that someone would want to use a dot and explicit paths in the same command.

But I could change it to merge the expansion of a dot in any position with the rest of the list if we would rather want that behaviour. I thought a bit about it when implementing it and found that solution to be a bit strange.

maxz:
I changed the behaviour to fix those points. The described scenarios do not make much sense to me, but I guess it would cause more requests and trouble to deny people from using it that way than to just allow it.

JustAnotherArchivist:
I suppose I wouldn't mind much if it was only supported when the dot is the only file parameter. But ignoring the others etc. is weird.

The new code would still change the order in which files are processed if the dot isn't the first file, which might lead to confusion but might be acceptable. Otherwise:

```python3
while '.' in local_file:
	i = local_file.index('.')
	local_file[i : i+1] = os.listdir('.')
```

(Which would also handle the weird edge case of someone specifying the current directory twice.)

maxz:
Well, specifying a dot together with files is already an edge case in my opinion.
I don't think that allowing for the same files to be twice in the list is sensible, especially if `--checksum` is not specified.
We should really protect the users from themselves in that case.

I could see the merit of preserving the relative order of the first period and dropping any other periods (basically changing the while in your example to an if and deleting additional periods).
But since the period is already expanded to not explicitly mentioned components, I don't see real additional potential for confusion when not preserving the order.

JustAnotherArchivist:
I don't see how that is any different from someone accidentally doing `ia upload $identifier dir dir`. I would *expect* it to upload things twice in that case (cf. `tar -c file file` producing output containing the file twice, `wget URL URL` downloading it twice, etc., none of which make much sense but might be reasonable in weird circumstances). A generic duplicate detection might be worthwhile, but it should definitely emit a warning in my opinion.

maxz:
Yes, it would probably be best to skip duplicate pathes.

I don't see any merit to uploading the same file twice as part of the same operation. I can't think of any sensible case for allowing this.

jjjake:
@maxz looking good to me, is this ready to merge?

maxz:
Yes, @jjjake.

PR #538, opened 2022-06-28
host could be same as in get_auth_config

Harsh-br0:
Raises a weird invalid host error without default arg


jimman2003:
https://github.com/internetarchive/openlibrary-client/issues/322 is caused by this..

jimman2003 (approved):
I think this will cover the default case quite well.. I tested this by setting the host argument when calling ia.configure , and that works

jjjake:
Any thoughts/feedback on this @cclauss?

cclauss (approved):
Seems like the correct default to me.

PR #541, opened 2022-07-28
Fixes memoryview error when running `Item.upload` with `StringIO` input and `verbose=True`

varun-iyer:
## The Bug
When running, for example:
```python
import io
import internetarchive as ia

ia.upload("ia_bug_test", {"test.txt": StringIO("test")}, access_key="", secret_key="", verbose=True)
```
instead of the expected message:
```
 uploading test.txt: 100%|███████████████████████████████████████████████████| 1/1 [00:00<00:00, 10.30MiB/s]
 ```
the Python output is a long error ending in:
 ```python
 TypeError: memoryview: a bytes-like object is required, not 'str'
```
(the entire error is pasted at the end of this message). This only occurs when the input is a `StringIO` object and `verbose=True`.

This error **is not reproducible in the test suite** because it depends on an interaction between `internetarchive` and the internals of `http`. `http` checks to see if body input is a `TextIOBase` object. If it is a `TextIOBase`, `http` encodes the string output of the `TextIOBase` into bytes. However, only when `verbose=True`, `internetarchive` wraps the `TextIOBase` in a `tqdm` iterable and then `IterableToFileAdapter`, which hides the information that, internally, the data from this object is from a `TextIOBase`. This causes `http`'s check to return `False`, so `http` does not encode the strings into bytes, which eventually causes this memoryview error.

## The Fix
The proposed fix to this issue makes a small modification to `IterableToFileAdapter` which automatically `encode`s data as it is read out. The `pre_encode` flag is set to enable this feature when we observe a `TextIOBase`, but it could be expanded to other situations where we have an input that `http` won't recognize and encode for us.

## Appendix
```
 uploading test.txt:   0%|                                                           | 0/1 [00:00<?, ?MiB/s]Traceback (most recent call last):
  File "/home/varun/harvard-lil/internetarchive/internetarchive/session.py", line 540, in send
    reraise_modify(e, e.request.url, prepend=False)  # type: ignore
AttributeError: 'TypeError' object has no attribute 'request'

During handling of the above exception, another exception occurred:

  File "/home/varun/harvard-lil/internetarchive/../perma/ia_test.py", line 7, in <module>
    internetarchive.upload("perma_cc_mvarun_bulktest", {"test.txt": StringIO("test")}, access_key=ACCESS, secret_key=SECRET, verbose=True)
  File "/home/varun/harvard-lil/internetarchive/internetarchive/api.py", line 277, in upload
    return item.upload(
  File "/home/varun/harvard-lil/internetarchive/internetarchive/item.py", line 1190, in upload
    resp = self.upload_file(body,
  File "/home/varun/harvard-lil/internetarchive/internetarchive/item.py", line 1006, in upload_file
    response = self.session.send(prepared_request,
  File "/home/varun/harvard-lil/internetarchive/internetarchive/session.py", line 543, in send
    raise e
  File "/home/varun/harvard-lil/internetarchive/internetarchive/session.py", line 537, in send
    r = super().send(request, **kwargs)
  File "/home/varun/harvard-lil/internetarchive/lib/python3.10/site-packages/requests/sessions.py", line 701, in send
    r = adapter.send(request, **kwargs)
  File "/home/varun/harvard-lil/internetarchive/lib/python3.10/site-packages/requests/adapters.py", line 489, in send
    resp = conn.urlopen(
  File "/home/varun/harvard-lil/internetarchive/lib/python3.10/site-packages/urllib3/connectionpool.py", line 703, in urlopen
    httplib_response = self._make_request(
  File "/home/varun/harvard-lil/internetarchive/lib/python3.10/site-packages/urllib3/connectionpool.py", line 398, in _make_request
    conn.request(method, url, **httplib_request_kw)
  File "/home/varun/harvard-lil/internetarchive/lib/python3.10/site-packages/urllib3/connection.py", line 239, in request
    super(HTTPConnection, self).request(method, url, body=body, headers=headers)
  File "/usr/lib/python3.10/http/client.py", line 1282, in request
    self._send_request(method, url, body, headers, encode_chunked)
  File "/usr/lib/python3.10/http/client.py", line 1328, in _send_request
    self.endheaders(body, encode_chunked=encode_chunked)
  File "/usr/lib/python3.10/http/client.py", line 1277, in endheaders
    self._send_output(message_body, encode_chunked=encode_chunked)
  File "/usr/lib/python3.10/http/client.py", line 1076, in _send_output
    self.send(chunk)
  File "/usr/lib/python3.10/http/client.py", line 1002, in send
    self.sock.sendall(d)
  File "/usr/lib/python3.10/ssl.py", line 1234, in sendall
    with memoryview(data) as view, view.cast("B") as byte_view:
TypeError: memoryview: a bytes-like object is required, not 'str'
```

### DIFF
--- a/internetarchive/item.py
+++ b/internetarchive/item.py
@@ -25,6 +25,7 @@ internetarchive.item
 """
 from __future__ import annotations
 
+import io
 import math
 import os
 import sys
@@ -953,11 +954,20 @@ class Item(BaseItem):
                     expected_size = math.ceil(size / chunk_size)
                     chunks = chunk_generator(body, chunk_size)
                     progress_generator = tqdm(chunks,
-                                              desc=f' uploading {key}',
-                                              dynamic_ncols=True,
-                                              total=expected_size,
-                                              unit='MiB')
-                    data = IterableToFileAdapter(progress_generator, size)
+                                          desc=f' uploading {key}',
+                                          dynamic_ncols=True,
+                                          total=expected_size,
+                                          unit='MiB')
+                    data = None
+                    # pre_encode is needed because http doesn't know that it
+                    # needs to encode a TextIO object when it's wrapped
+                    # in the Iterator from tqdm.
+                    # So, this FileAdapter provides pre-encoded output
+                    data = IterableToFileAdapter(
+                        progress_generator,
+                        size,
+                        pre_encode=isinstance(body, io.TextIOBase)
+                    )
                 except Exception:
                     print(f' uploading {key}', file=sys.stderr)
                     data = body
--- a/internetarchive/item.py
+++ b/internetarchive/item.py
@@ -954,10 +954,10 @@ class Item(BaseItem):
                     expected_size = math.ceil(size / chunk_size)
                     chunks = chunk_generator(body, chunk_size)
                     progress_generator = tqdm(chunks,
-                                          desc=f' uploading {key}',
-                                          dynamic_ncols=True,
-                                          total=expected_size,
-                                          unit='MiB')
+                                              desc=f' uploading {key}',
+                                              dynamic_ncols=True,
+                                              total=expected_size,
+                                              unit='MiB')
                     data = None
                     # pre_encode is needed because http doesn't know that it
                     # needs to encode a TextIO object when it's wrapped
--- a/internetarchive/session.py
+++ b/internetarchive/session.py
@@ -121,7 +121,8 @@ class ArchiveSession(requests.sessions.Session):
         self.secret_key: str = self.config.get('s3', {}).get('secret')
         self.http_adapter_kwargs: MutableMapping = http_adapter_kwargs or {}
 
-        self.headers = default_headers()
+        # mypy doesn't like 'headers: Dict[Union[bytes, str]] = Dict[str]'
+        self.headers = default_headers()  # type: ignore
         self.headers.update({'User-Agent': self._get_user_agent_string()})
         self.headers.update({'Connection': 'close'})
 
--- a/internetarchive/utils.py
+++ b/internetarchive/utils.py
@@ -134,11 +134,18 @@ def suppress_keyboard_interrupt_message() -> None:
 
 
 class IterableToFileAdapter:
-    def __init__(self, iterable, size: int):
+    def __init__(self, iterable, size: int, pre_encode: bool = False):
         self.iterator = iter(iterable)
         self.length = size
+        # pre_encode is needed because http doesn't know that it
+        # needs to encode a TextIO object when it's wrapped
+        # in the Iterator from tqdm.
+        # So, this FileAdapter provides pre-encoded output
+        self.pre_encode = pre_encode
 
     def read(self, size: int = -1):  # TBD: add buffer for `len(data) > size` case
+        if self.pre_encode:
+            return next(self.iterator, '').encode("iso-8859-1")
         return next(self.iterator, b'')
 
     def __len__(self) -> int:
--- a/internetarchive/utils.py
+++ b/internetarchive/utils.py
@@ -145,6 +145,11 @@ class IterableToFileAdapter:
 
     def read(self, size: int = -1):  # TBD: add buffer for `len(data) > size` case
         if self.pre_encode:
+            # this adapter is intended to emulate the encoding that is usually
+            # done by the http lib.
+            # As of 2022, iso-8859-1 encoding is used to meet the HTTP standard,
+            # see in the cpython repo (https://github.com/python/cpython
+            # Lib/http/client.py lines 246; 1340; or grep 'iso-8859-1'
             return next(self.iterator, '').encode("iso-8859-1")
         return next(self.iterator, b'')
 


jjjake:
Thank you @varun-iyer! This looks good to me besides the failed checks above.

@cclauss Any feedback besides fixing the failed checks?

cclauss:
For `bytes`, I would normally use `io.BytesIO()` instead of `io.StringIO`.

varun-iyer:
@cclauss, I'm not sure I fully understand your comment --- but the basic problem here is that we have a `StringIO` input that, at some point, needs to be converted into bytes since `http` won't do that for us automatically because of the `IterableToFileAdapter` wrapper.

The current PR does this conversion as bytes are read out of the adapter. However, it could possibly be done earlier in the process by converting the `StringIO` into a `BytesIO` object before it goes into the `IterableToFileAdapter`. Is this what you are suggesting?

cclauss:
my_string.encode() --> bytes
my_bytes.decode() --> str

varun-iyer:
@cclauss sorry --- I’m still not understanding what change you’re proposing to the pull request!
To be clear, I am not trying to upload raw bytes to IA, and StringIO input is a [documented feature](https://archive.org/services/docs/api/internetarchive/quickstart.html?highlight=stringio#uploading) that works as expected when `verbose=False`.

cclauss (approved):
RSLGTM

cclauss:
@jjjake Please approve the GitHub Actions so the tests are run.

varun-iyer:
I'm currently able to reproduce the two `mypy` test failures on `jjjake/internetarchive:master` --- maybe some upstream change?

It looks like an explicit type specification fixed the issue, but I don't have much experience with mypy. In any case, tests are passing now!

jjjake:
@varun-iyer Looks like we're having issues with the `CaseInsensitiveDict` import. It does look like we can import it from `requests` though:

```
>>> from requests.structures import CaseInsensitiveDict
```


Besides that, looking good! Thanks again.

varun-iyer:
Should I add the CaseInsensitiveDict dependency here, or should I just mark that line to be ignored by mypy?

varun-iyer:
Fixed! thanks for the suggestion @jjjake!

jjjake:
Thanks @varun-iyer! I'll merge after the tests finish.

jjjake:
Shoot, looks like it still didn't like that. Perhaps ignoring it, as you suggested, is the way to go. I'm still familiarizing myself with mypy as well, so I'm curious if @cclauss has any feedback here. Thanks again, and my apologies for the issues!

varun-iyer:
Arg, I wasn't running into this with pre-commit yesterday. I tried switching to `typing.cast` instead, that seems to do the trick locally.

varun-iyer:
Looks like this is a problem with the change in typing syntax with Python 3.9/3.10 --- I'm thinking of giving up on whack-a-mole and just having mypy ignore this line, if that works for you. The problem anyway is that the variable is typed to `Dict[Union[str, bytes]]` and `mypy` is losing it when trying to assign a `Dict[str]` which is always a subset of `Dict[Union[str, bytes]]`...

jjjake:
@varun-iyer Yes, having mypy ignore this line works for me. Thanks a lot for the contribution and research into the problem, I appreciate it!

varun-iyer:
@jjjake happy to help! Changed the line to `# type: ignore`

cclauss:
I ignored that line also in #542

internetarchive/session.py
```diff
@@ -34,7 +34,7 @@
 import platform
 import sys
 import warnings
-from typing import Iterable, Mapping, MutableMapping
+from typing import Iterable, Mapping, MutableMapping, Union, cast
```
JustAnotherArchivist, line 37: Neither `Union` nor `cast` are used below (anymore), so they shouldn't be imported either.

internetarchive/utils.py
```diff
@@ -134,11 +134,18 @@ def new_hook(type, value, traceback):
...
         self.length = size
+        # pre_encode is needed because http doesn't know that it
+        # needs to encode a TextIO object when it's wrapped
+        # in the Iterator from tqdm.
+        # So, this FileAdapter provides pre-encoded output
+        self.pre_encode = pre_encode
 
     def read(self, size: int = -1):  # TBD: add buffer for `len(data) > size` case
+        if self.pre_encode:
+            return next(self.iterator, '').encode("iso-8859-1")
```
JustAnotherArchivist, line 148: Do we really want to go with an ancient, obsolete encoding here rather than UTF-8?

internetarchive/item.py
```diff
@@ -953,11 +954,20 @@ def _build_request():
...
                     progress_generator = tqdm(chunks,
-                                              desc=f' uploading {key}',
-                                              dynamic_ncols=True,
-                                              total=expected_size,
-                                              unit='MiB')
-                    data = IterableToFileAdapter(progress_generator, size)
+                                          desc=f' uploading {key}',
+                                          dynamic_ncols=True,
+                                          total=expected_size,
+                                          unit='MiB')
```
JustAnotherArchivist, line 960: Style nit: leave the indentation of the `tqdm` arguments as it was.

internetarchive/session.py
cclauss: Enable the appropriate flake8 is a separate pr.

internetarchive/item.py
cclauss: Let’s have machines do code formatting so people do not have to.  #532

internetarchive/utils.py
varun-iyer: This was chosen to match the encoding performed by the HTTP library that this PR emulates. I can test it out with UTF-8, but I have more trust that matching the library exactly will best handle all possible cases.

varun-iyer:
@jjjake @JustAnotherArchivist addressed the style issues, left the encoding alone for now.  Let me know what your thoughts are.

internetarchive/utils.py
varun-iyer: There are a few references in the http lib suggesting that ISO 8859-1 is required in parts of the http standard, see:

https://github.com/python/cpython/blob/e87ada48a9e5d9d03f9759138869216df0d7383a/Lib/http/client.py#L246
https://github.com/python/cpython/blob/e87ada48a9e5d9d03f9759138869216df0d7383a/Lib/http/client.py#L1340

so I think it is best to keep it this way, I will add a comment noting the reason though.

jjjake:
Thanks again @varun-iyer!

Issue #544, opened 2022-08-10
Add option to exclude files that match a glob pattern


thetaiter:
I'd like to be able to include only files that match a glob pattern, except those matching some exclude pattern. I was unable to find a way to do that with the existing `ia` tool.

jjjake:
fixed in #545

PR #545, opened 2022-08-10
Add ability to exclude files that match a glob pattern

thetaiter:
This is my attempt at adding [this feature](https://github.com/jjjake/internetarchive/issues/544) myself. It appears to work, let me know if there is anything else I can do to make this change merge-worthy :)


thetaiter:
I think I've gotten everything in order as far as the tests go at this point. Ready for review.

internetarchive/__version__.py
```diff
@@ -1 +1 @@
-__version__ = '3.0.2'
+__version__ = '3.0.3'
```
maxz, line 1: You should not change the version. That's what Jake does when preparing a new release.

maxz (approved):
Apart from the version bump, the pull request looks good.

internetarchive/__version__.py
thetaiter: @maxz Sorry about that. This change has been reverted. Thanks!

jjjake:
@thetaiter This is really great, and has been a long requested feature! Thanks so much for the contribution, we really appreciate it!

Issue #490, opened 2022-02-19
Continue to upload files if one file doesn't exist


eubrunosilva:
Hi
I'm uploading thru csv and I have some files in list that doesn't exist in  my hdd (were renamed meanwhile).
It's possible to IA to continue to process the list (giving the error) and not stop the upload.
Thanks

![image](https://user-images.githubusercontent.com/16777233/154797700-5072dae7-347d-4648-9637-2f13d3251f44.png)

jjjake:
Thanks for the feedback @eubrunosilva!

I can see how this would be helpful. I think it should continue on with the rest of the upload, printing an error message for any missing files, and exiting with a non-zero status code if any files were missing.

eubrunosilva:
yes... if this could be implemented, I would appreciate

jjjake:
Reopening due to having to revert the fix @mshemuni provided due to a few new bugs being introduced (more details [here](https://github.com/jjjake/internetarchive/commit/4a234de3c1a32417b9a30f3a1b619ec1c0b1dd15#commitcomment-92349736)).

PR #551, opened 2022-10-25
Issue $490

mshemuni:
A function was added to exclude file paths that do not exist. https://github.com/jjjake/internetarchive/issues/490

### DIFF
--- a/internetarchive/cli/ia_upload.py
+++ b/internetarchive/cli/ia_upload.py
@@ -86,21 +86,20 @@ from internetarchive.utils import (
 )
 
 
-def _check_files(files):
+def _check_files_exist(files):
     return [
-        file
-        for file in files
+        file for file in files
         if Path(file).exists()
     ]
 
 
 def _upload_files(item, files, upload_kwargs, prev_identifier=None, archive_session=None):
     """Helper function for calling :meth:`Item.upload`"""
-    # Exclude files that do not exist.
-    files = _check_files(files)
+    # Exclude files that do not exist. (Make a copy of the list)
+    files = _check_files_exist(files)
 
     # Check if the list has any element.
-    if len(files) < 1:
+    if not files:
         raise FileNotFoundError("No valid file was found. Check your paths.")
 
     responses = []


internetarchive/cli/ia_upload.py
```diff
@@ -85,8 +86,23 @@
 )
 
 
+def _check_files(files):
```
maxz, line 89: Should have a better name e. g. `_check_files_exist` or something else which makes it clear that it merely checks whether a path leads to an actual file.

internetarchive/cli/ia_upload.py
```diff
@@ -85,8 +86,23 @@
...
+    return [
+        file
```
maxz, line 91: This is very strangely formatted and should at least have `file for file in files` on a single line to make it more apparent what is happening.

internetarchive/cli/ia_upload.py
```diff
@@ -85,8 +86,23 @@
...
+    ]
+
+
 def _upload_files(item, files, upload_kwargs, prev_identifier=None, archive_session=None):
     """Helper function for calling :meth:`Item.upload`"""
+    # Exclude files that do not exist.
+    files = _check_files(files)
+
+    # Check if the list has any element.
+    if len(files) < 1:
```
maxz, line 103: Use the idiomatic `if not files:` instead.

maxz (changes requested):
Thank you for the contribution.
The general solution is alright but you should adjust some minor details to ensure that everything remains readable.

A cherry on top would be if you could change the first line of your commit to `Fix issue #490` and the last line to `Closes #490` so that the issue will be automatically close when merging this patch.
It might be worth mentioning that there is a typo in your second line and should be "do" instead of "does" and "exist" instead of "exits".

internetarchive/cli/ia_upload.py
mshemuni: Ye I saw them now. Thank you for your answer. Another problem that I noticed is that `files` is an immutable data type, so altering it might not be a good idea.

internetarchive/cli/ia_upload.py
mshemuni: I am not really familiar with the whole project so I'm not sure if anything other than a list of strings can be passed to the function or not. That's why I did not trust truthy/falsy values. But since you mentioned it, there can't be a problem.

mshemuni:
Before committing fixed I'm going to mention since `files` is expected to be a `list` (a mutable) I shouldn't alter it. So basically I'm going to make a copy of it.

maxz:
You are not altering it. The generator expression effectively creates a copy (unless complex objects are involved which they are not in this case).

maxz (approved):
The commit message could have been better but lets not nitpick too much.
I'm mentioning it so that you can improve it next time: When you have both in the first line anyways, then `Fixes #490` would have been enough, that would also close the issue. There are certain other [keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) that would also work. 
Afterwards you also have both `Fix issue jjjake#490` and `Closes jjjake#490` again in the message, which is unnecessary.

jjjake:
Thanks for the contribution @mshemuni and for the review once again @maxz!

maxz:
Jake reverted this change in commit `d6a820e1cb7adb45953af137519c513e9d331f34` for understandable reasons. It does not work with files from stdin which have no filename and it does not work with files which are supposed to have another file name on the server.

The remote-name part could be fixed by checking whether `files` within `_check_files` is a dictionary and to then check whether the value of the element within that dictionary (which is the local filename) exists on the filesystem. (e. g. `files = {args['--remote-name']: local_file}` or resolved as `{"my-fancy-remote-filename.jpg": "local-file.jpg"}`)
The stdin part could be fixed by checking whether `files` is a `tempfile.TemporaryFile` and skipping the `.exists()` check in that case.

PR #572, opened 2022-12-28
fix directory race conditions

lunixbochs:
`files.py` has a few TOCTOU race conditions wrt checking if a file or directory exists before performing an operation. I hit one of these during a parallel download.

It's safer to just try the operation (e.g. mkdir) and ignore the FileNotFound or FileExists exception, and this approach requires fewer IO syscalls too.

### DIFF
--- a/internetarchive/files.py
+++ b/internetarchive/files.py
@@ -199,8 +199,11 @@ class File(BaseFile):
         file_path = file_path or self.name
 
         if destdir:
-            if not os.path.exists(destdir) and return_responses is not True:
-                os.mkdir(destdir)
+            if return_responses is not True:
+                try:
+                    os.mkdir(destdir)
+                except FileExistsError:
+                    pass
             if os.path.isfile(destdir):
                 raise OSError(f'{destdir} is not a directory!')
             file_path = os.path.join(destdir, file_path)
@@ -233,12 +236,10 @@ class File(BaseFile):
                     return
 
         parent_dir = os.path.dirname(file_path)
-        if parent_dir != '' \
-                and not os.path.exists(parent_dir) \
-                and return_responses is not True:
-            os.makedirs(parent_dir)
-
         try:
+            if parent_dir != '' and return_responses is not True:
+                os.makedirs(parent_dir, exist_ok=True)
+
             response = self.item.session.get(self.url,
                                              stream=True,
                                              timeout=12,
@@ -272,8 +273,10 @@ class File(BaseFile):
         except (RetryError, HTTPError, ConnectTimeout, OSError, ReadTimeout) as exc:
             msg = f'error downloading file {file_path}, exception raised: {exc}'
             log.error(msg)
-            if os.path.exists(file_path):
+            try:
                 os.remove(file_path)
+            except OSError:
+                pass
             if verbose:
                 print(f' {msg}', file=sys.stderr)
             if ignore_errors:


internetarchive/files.py
```diff
@@ -233,10 +236,8 @@ def download(self, file_path=None, verbose=None, ignore_existing=None,
                     return
 
         parent_dir = os.path.dirname(file_path)
-        if parent_dir != '' \
-                and not os.path.exists(parent_dir) \
-                and return_responses is not True:
-            os.makedirs(parent_dir)
+        if parent_dir != '' and return_responses is not True:
+            os.makedirs(parent_dir, exist_ok=True)
```
maxz, line 240: This will throw an exception if a file by that intended directory name exists.

internetarchive/files.py
```diff
@@ -272,8 +273,10 @@ def download(self, file_path=None, verbose=None, ignore_existing=None,
         except (RetryError, HTTPError, ConnectTimeout, OSError, ReadTimeout) as exc:
             msg = f'error downloading file {file_path}, exception raised: {exc}'
             log.error(msg)
-            if os.path.exists(file_path):
-                os.remove(file_path)
+            try:
+                os.remove(file_path.encode('utf-8'))
```
maxz, line 277: I cannot produce a case where explicitly encoding as UTF-8 would be necessary or sensible. It would probably even be harmful on systems which do not use UTF-8.

maxz (changes requested):
Thank you for the contribution.

You introduced an error which you should fix and remove the explicit encoding.

internetarchive/files.py
lunixbochs: There are a bunch of other file path utf8 encodes in this file, I added this one because I did run into a file downloaded from the archive where this line threw an exception without the encoding. So I vote either leave it, or remove all the other ones too.

internetarchive/files.py
lunixbochs: I don't think that's a new issue.

The old code would have thrown an exception as well, if:
- Any of the directory components is a file, but the final path does not exist.
- If the parent directory exists as a file, it will probably throw here: https://github.com/jjjake/internetarchive/blob/master/internetarchive/files.py#L264

maxz: I cannot follow your first point.

But it is true that the issue existed before. Still it now throws on the `os.makedirs` call where it before most likely threw on your marked line. I'm not sure what the ramifications of this are and whether that exception should be silenced to retain behaviour identical to the current program.

internetarchive/files.py
maxz: Could you please provide a link to the item and file that produced the error?

I tested it with various characters and could not cause an exception.
If it is really necessary then we should change it to encode to `sys.getfilesystemencoding()` everywhere and even add it in the places where it is currently missing.

lunixbochs: I was wrong on this one, it just took a while to resurface because I'm working with such a large dataset. I made an assumption because it was a weird unicode filename and there were other utf8 encodes in the same file, but the filename was actually just way too long, and unicode was masking that.

My sys.getfilesystemencoding() is actually utf8 already, which means the encoding is doing nothing. Mild vote to remove the other encodes.

internetarchive/files.py
lunixbochs: If your parent dir is `a/b/c/d`, and `a/b` exists as a file, `os.path.exists("a/b/c/d") == False`, but `os.makedirs` will still error out due to not being able to make a directory under a file path - so "path component is a file during mkdir" was already an issue, the `exists()` check only ruled it out for the final component.

I also confirmed that you get `NotADirectoryError` for both `os.makedirs("a/b/c/d")`, and for `open("a/b/c/d", "w")`, so the exception line changes but it throws the same exception. IMO this is not a regression.

lunixbochs: I pushed a new version that catches OSError instead of FileNotFoundError for the os.remove, and moved this makedirs into the try block so it will be caught in the same way as previous open() error

maxz (approved):

jjjake:
Thank you @lunixbochs, I appreciate it! And thanks for the review @maxz! :)

PR #583, opened 2023-04-03
Fix TypeError when iterating over 0 search results

DuncanDHall:
Scrape can result in `j` where `count: 0, total: None` resulting in:
`TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'`

### DIFF
--- a/internetarchive/search.py
+++ b/internetarchive/search.py
@@ -153,7 +153,7 @@ class Search:
             if j.get('error'):
                 yield j
             if not num_found:
-                num_found = int(j['total'])
+                num_found = int(j['total'] or 0)
             if not self._num_found:
                 self._num_found = num_found
             self._handle_scrape_error(j)


internetarchive/search.py
```diff
@@ -153,7 +153,7 @@ def _scrape(self):
             if j.get('error'):
                 yield j
             if not num_found:
-                num_found = int(j['total'])
+                num_found = int(j['total'] or 0)
             if not self._num_found:
                 self._num_found = num_found
```
cclauss, line 158: It would be awesome if we had an empty search test case that failed with the current code and passed with the proposed code.

The GitHub Actions tests have been updated so another push to the pull request should pass all GitHub Actions.

Please remove line 155 and then...
```suggestion
            num_found = num_found or int(j['total'] or 0)
            self._num_found = self._num_found or num_found
```

PR #605, opened 2023-08-20
Add --start-idx=<n> option to start download directly from item <n>

Russtopia:
Quick and dirty addition of the `--start-idx=<n>` option for cli download; this speeds up resumption of incomplete collection downloads immensely, directing the iteration loop to skip (n-1) entries without checking them against locally-downloaded items.


Russtopia:
I will update ASAP to add to the AUTHORS.rst and look at adding tests. Thanks. (To comply with the contributing guidelines).

internetarchive/cli/ia_download.py
```diff
@@ -33,6 +33,8 @@
     -R, --retries=<retries>                  Set number of retries to <retries> [default: 5].
     -I, --itemlist=<file>                    Download items from a specified file. Itemlists should
                                              be a plain text file with one identifier per line.
+    -n, --start-idx=<n>                      Start immediately at item <n>
```
maxz, line 36: Change the long argument name to --start-index. The short from `-n` should probably be removed.

internetarchive/cli/ia_download.py
```diff
@@ -33,6 +33,8 @@
...
+    -m, --end-idx=<m>                        End download after item <m>
```
maxz, line 37: Change the long argument name to --end-index. The short from `-m` should probably be removed.

internetarchive/cli/ia_download.py
```diff
@@ -128,6 +132,18 @@ def main(argv, session: ArchiveSession) -> None:
         print(f'{exc}\n{printable_usage(__doc__)}', file=sys.stderr)
         sys.exit(1)
 
+    if args['--start-idx']:
+      start_idx = int(args['--start-idx'])-1
```
maxz, line 136: Add spaces around all operators `int(args['--start-index']) - 1`. Also use four spaces for indentation as used everywhere else in the code ...
And if it should no be clear from the argument remark, also use start_index for the variable name instead of using idx.

internetarchive/cli/ia_download.py
```diff
@@ -128,6 +132,18 @@ def main(argv, session: ArchiveSession) -> None:
...
+      print(f'Starting download at collection item {start_idx+1}')
```
maxz, line 137: Again, spaces.

internetarchive/cli/ia_download.py
```diff
@@ -167,53 +183,59 @@ def main(argv, session: ArchiveSession) -> None:
 
     errors = []
     for i, identifier in enumerate(ids):
-        try:
-            identifier = identifier.strip()
-        except AttributeError:
-            identifier = identifier.get('identifier')
-        if total_ids > 1:
-            item_index = f'{i + 1}/{total_ids}'
+        if end_idx != None and end_idx == i:
```
maxz, line 186: None is a singleton. Use `is not None` for the comparison. In this case `end_index == i` would also suffice as condition.

internetarchive/cli/ia_download.py
```diff
@@ -167,53 +183,59 @@ def main(argv, session: ArchiveSession) -> None:
...
-            params=args['--parameters'],
-            ignore_history_dir=ignore_history_dir,
-            source=args['--source'],
-            exclude_source=args['--exclude-source'],
-            stdout=args['--stdout'],
-            timeout=args['--timeout'],
-        )
-        if _errors:
-            errors.append(_errors)
+          try:
```
maxz, line 191: The indentation in this whole section is wrong.

internetarchive/cli/ia_download.py
```diff
@@ -167,53 +183,59 @@ def main(argv, session: ArchiveSession) -> None:
...
+              params=args['--parameters'],
+              ignore_history_dir=ignore_history_dir,
+              source=args['--source'],
+              exclude_source=args['--exclude-source'],
+              stdout=args['--stdout'],
+              timeout=args['--timeout'],
+          )
+          if _errors:
+              errors.append(_errors)
+        ##endif (start_idx)
```
maxz, line 238: Remove this superfluous line.

maxz (changes requested):
Thank you for the patch.

You should really run the tests locally against your changes. See https://github.com/jjjake/internetarchive/blob/master/CONTRIBUTING.rst

Your indentation is often wrong and you should not abbreviate index.
Also you should add spaces around operators.

If you had run the tests, Black would have complained about all those things except for the index part which it can't sensibly catch.

Using the index does not make too much sense in my opinion, because the collection could change at any time and using identifiers to mark the beginning and end would probably lead to more consistent results.

Russtopia:
OK, I will look at making all of those changes and re-submit.

In regards to marking beginning and end -- what identifier do you think might be best to use? I know little to nothing about how the IA uses metadata and so on; perhaps there's a specific one that will be constant even as a collection is updated?

(I have already noticed a collection I was testing this on has grown since I began; but so far the older items hadn't moved around so resuming at a specific item number seemed to result in the same one each time.)

maxz:
As far as I know there is no marker which remains always constant in regard to the sequence. But I also think that it is close to impossible to cover the edge cases here. (e. g. an existing item could also just be replaced)

I would think that using the item's identifier (i. e. the "name" of the item) would at least give a somewhat more reliable outcome when splitting a sequence. But every approach has its own issues.
E. g. you could say `--start-at=item-abc-1 --end-at=item-xyz-555`. But while the start and end item would remain constant with this scheme, you could still miss items. So it's not a perfect solution either.

I guess your current approach is easier to use for your use case.

jjjake:
What if you just used `--search-parameters` to set a sort like so:

```ia download --search 'frogs' --search-parameters='sorts=publicdate desc'```

And then when you resumed, you could do something like:

```ia download --search 'frogs AND publicdate:[YYYY-MM-DD TO *]' --search-parameters='sorts=publicdate desc'```

Alternatively, you could use `--itemlist` (or an itemlist [used with GNU Parallel](https://archive.org/developers/internetarchive/parallel.html) and `ia download`) and modify your itemlist as necessary on resume.

PR #606, opened 2023-08-23
Patch search error

DuncanDHall:
Currently key error is thrown when elastic search returns no 'response' key in its json when we try to grab the num_found.


jjjake:
Thanks @DuncanDHall!

Issue #613, opened 2023-10-21
[Bug] No module named 'pkg_resources'


ego-lay-atman-bay:
I just installed this package in python 3.12 by running
```
pip install internet archive
```
It was able to install just fine, but when I ran `ia`, it threw this error
```
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "C:\Python\Python312\Scripts\ia.exe\__main__.py", line 4, in <module>
  File "C:\Python\Python312\Lib\site-packages\internetarchive\cli\__init__.py", line 26, in <module>
    from internetarchive.cli import (
  File "C:\Python\Python312\Lib\site-packages\internetarchive\cli\ia.py", line 67, in <module>
    from pkg_resources import DistributionNotFound, iter_entry_points
ModuleNotFoundError: No module named 'pkg_resources'
```

I downloaded the repo, and opened it in vscode. After I installed the `pex-requirements.txt`, vscode said that `pkg_resources` was found.

I think this might be a problem with python 3.12, because I was able to get it installed and working just fine with python 3.11.

jjjake:
Thanks for the report @ego-lay-atman-bay. I am able to replicate this. Does this fix it for you (it did for me)?:

```pip install setuptools```

I'll look into this, thanks again for the report.

clydecbrown:
> `pip install setuptools`

This fixed it for me.

macOS 12.7.1
Homebrew Python 3.12.1
internetarchive 3.6.0

ego-lay-atman-bay:
Oh yeah, that worked for me too. Should this stay open for other people who run into this issue?

JustAnotherArchivist:
Python packaging has changed (read: improved) a fair bit in the past few years. `setuptools` used to be the one and only way to do anything with packages. That's not the case anymore. Packages can be installed with something else (if they support it), and it might not be present at runtime.

As long as `pkg_resources` is required by the CLI, `internetarchive` should depend on `setuptools` via `install_requires`. The `build-backend` should also be specified in `pyproject.toml`, though that's unrelated to this issue.

PR #621, opened 2024-01-09
Switch to importlib-metadata to drop deprecated pkg_resources

FantasqueX:
According to https://setuptools.pypa.io/en/latest/pkg_resources.html, pkg_resources has been deprecated and importlib-metadata is recommended. `DistributionNotFound` only can be thrown from `find_plugins()` which is not used by ia. Tested with plugin https://github.com/JesseWeinstein/ia_recent.

Closes: https://github.com/jjjake/internetarchive/issues/613

### DIFF
--- a/internetarchive/cli/ia.py
+++ b/internetarchive/cli/ia.py
@@ -64,7 +64,10 @@ import os
 import sys
 
 from docopt import docopt, printable_usage
-from pkg_resources import DistributionNotFound, iter_entry_points
+if sys.version_info < (3, 10):
+  from importlib_metadata import entry_points
+else:
+  from importlib.metadata import entry_points
 from schema import Or, Schema, SchemaError  # type: ignore[import]
 
 from internetarchive import __version__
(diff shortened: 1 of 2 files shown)


maxz:
That's a nice and rather simple change. It seems like the API is mostly compatible and it is a good further step in modernizing `ia`.

jjjake:
Thanks for the contribution @FantasqueX, and thanks for reviewing @maxz!

Looks good to me, I just want to learn more about importlib before adding it as a requirement.

setup.cfg
```diff
@@ -26,6 +26,7 @@ packages =
     internetarchive.cli
 install_requires =
     docopt>=0.6.0,<0.7.0
+    importlib-metadata; python_version <= 3.7
```
branchvincent, line 29: fyi, the `group` keyword wasn't added until python 3.10

internetarchive/cli/ia.py
```diff
@@ -64,7 +64,7 @@
 import sys
 
 from docopt import docopt, printable_usage
-from pkg_resources import DistributionNotFound, iter_entry_points
+from importlib.metadata import entry_points
```
branchvincent, line 67: this needs a conditional import:
```suggestion
if sys.version_info < (3, 10):
  from importlib_metadata import entry_points
else:
  from importlib.metadata import entry_points
```

FantasqueX: Thanks for your suggestions. I've fixed the patch according to your suggestions.

cclauss (approved):

PR #624, opened 2024-02-02
Add support for JSON Patch test operations

jjjake:


MerlijnWajer:
I changed the loop to this to make insert / append-list / remove for lists work, at least in some light testing:

```
for p in patch:
    path_parts = p['path'].split('/')
    if not source_metadata.get(path_parts[1]):
        continue
    if len(path_parts) == 2:
        src_val = source_metadata.get(path_parts[-1])
        p_test = {'op': 'test', 'path': p['path'], 'value': src_val}
    else:
        src_val = source_metadata.get(path_parts[1], [])
        p_test = {'op': 'test', 'path': '/' + path_parts[1], 'value': src_val}

    patch_tests.append(p_test)
```

jjjake:
Thanks @MerlijnWajer! I just pushed a patch with your fix.

PR #627, opened 2024-02-13
Bump test suite to current versions

JustAnotherArchivist:
CPython 3.7, PyPy 3.7, and PyPy 3.8 are EOL and unsupported.

The test on 3.12 currently fails. This is due to #613, and merging this PR should probably wait for #621 (or we should add a dependency on `setuptools` until that is resolved, as mentioned in <https://github.com/jjjake/internetarchive/issues/613#issuecomment-1872954045>).

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -7,7 +7,7 @@ jobs:
       fail-fast: false
       max-parallel: 1  # Avoid timeout errors
       matrix:
-        python: ['3.7', '3.8', '3.9', '3.10', '3.11', 'pypy-3.7', 'pypy-3.8', 'pypy-3.9']
+        python: ['3.8', '3.9', '3.10', '3.11', '3.12', 'pypy-3.9', 'pypy-3.10']
     steps:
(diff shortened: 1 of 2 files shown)


cclauss (approved):

cclauss (changes requested):
> we should add a dependency on `setuptools`

This is the correct solution because `distutils` was removed in Python 3.12 and the Python Packaging Authority recommends adding [`setuptools`](https://packaging.python.org/en/latest/key_projects/#distutils) to project dependencies.

JustAnotherArchivist:
Yeah, it's the correct solution – *if* we don't migrate away from `pkg_resources`. But the plan (#621) is to do so, and once that is merged, nothing here will depend on `setuptools` anymore at runtime. Thus the test on Python 3.12 will also work fine at that point (without any changes in this PR).

cclauss:
Yes.  #621 is a better approach.

JustAnotherArchivist:
Looks like this got obsoleted by e84edd1c although that did not remove the old PyPy versions. I'll create a new PR for that instead.

PR #628, opened 2024-02-13
Fix linting: disable mypy --install-types

JustAnotherArchivist:
`pip install .[all]` already installs the third-party hints, and `--install-types` prompts for confirmation, breaking the CI.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -80,7 +80,6 @@ ignore-words-list = alers
 [mypy]
 exclude = ^\.git/|^__pycache__/|^docs/source/conf.py$|^old/|^build/|^dist/|\.tox
 python_version = 3.9
-install_types = True
 pretty = True
 scripts_are_modules = True
 show_error_codes = True


JustAnotherArchivist:
Example of a linting crash on the mypy step from <https://github.com/jjjake/internetarchive/actions/runs/7892724829/job/21539842879>:

```
Installing missing stub packages:
/opt/hostedtoolcache/Python/3.12.2/x64/bin/python -m pip install types-requests

  File "/opt/hostedtoolcache/Python/3.12.2/x64/bin/mypy", line 8, in <module>
    sys.exit(console_entry())
             ^^^^^^^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.12.2/x64/lib/python3.12/site-packages/mypy/__main__.py", line 15, in console_entry
    main()
  File "mypy/main.py", line 133, in main
  File "mypy/main.py", line 1568, in install_types
EOFError: EOF when reading a line
Install? [yN] 
Error: Process completed with exit code 1.
```

I'm not sure why it thinks `types-requests` is missing; the `pip install .[all]` step does install it, as expected...

JustAnotherArchivist:
@cclauss Would you mind verifying whether I'm understanding things correctly here?

cclauss (approved):
Nice!

cclauss:
Please consider fixing the warning at the bottom right of https://github.com/jjjake/internetarchive/actions/runs/7893419664

This should be done in a separate PR like:
* jazzband/django-debug-toolbar#1876

JustAnotherArchivist:
Yeah, that was next on my list. Looks like the fix is to bump `actions/checkout` to version 4.

setup.cfg
```diff
@@ -80,7 +80,6 @@ ignore-words-list = alers
 [mypy]
 exclude = ^\.git/|^__pycache__/|^docs/source/conf.py$|^old/|^build/|^dist/|\.tox
 python_version = 3.9
-install_types = True
```
cclauss, line 83: Perhaps instead add `non-interactive = True`
* https://mypy.readthedocs.io/en/stable/command_line.html#miscellaneous

JustAnotherArchivist: Potentially. Although I think we'd ideally not have mypy install any type hint packages. I'm still puzzled why it even tries. That might indicate that something else is wrong.

cclauss: You can either tell it everything (constant human effort to align type hints) or allow it to guess.

JustAnotherArchivist: Yeah, but we already have the type hint packages in `.[types]`, and they only change when our dependencies do. Here, it tries to install `types-requests`, which is already getting installed (via `.[all]`), yet somehow mypy can't find it...

cclauss: What happens if you modify line 69 to exactly match https://pypi.org/project/types-requests ?

JustAnotherArchivist: Not sure what you mean. The version dependency on line 69 of `setup.cfg` matches the one on line 30 for `requests` (so that it should install matching versions), and we can see that it does install the latest PyPI version [here](https://github.com/jjjake/internetarchive/actions/runs/7893419666/job/21542016167#step:5:168).

Quoting for future reference since the Actions logs expire stupidly quickly:

```
Run pip install .[all]
<...>
Collecting requests<3.0.0,>=2.25.0 (from internetarchive==3.6.0)
  Downloading requests-2.31.0-py3-none-any.whl.metadata (4.6 kB)
<...>
Collecting types-requests<3.0.0,>=2.25.0 (from internetarchive==3.6.0)
  Downloading types_requests-2.31.0.20240125-py3-none-any.whl.metadata (1.8 kB)
<...>
Downloading requests-2.31.0-py3-none-any.whl (62 kB)
<...>
Downloading types_requests-2.31.0.20240125-py3-none-any.whl (14 kB)
<...>
Installing collected packages: <...>, types-requests, <...>, requests, <...>
Successfully installed <...> requests-2.31.0 <...> types-requests-2.31.0.20240125 <...>
```

cclauss (changes requested):
```
install_types = True
non_interactive = True
```

PR #631, opened 2024-02-17
Keep GitHub Actions up to date with GitHub's Dependabot

cclauss:
Fixes #630

Auto-generates GitHub Action upgrade pull requests like
* #630


JustAnotherArchivist:
NB, `install_types` vs `non_interactive` was discussed over in #628 (it indicates that something else might be wrong in the test env), and the PyPy bump here clashes with #627. I think this PR should remain limited to just the dependabot proposal.

cclauss:
Auto-generated:
* #634

PR #642, opened 2024-04-09
README.rst: add link to Repology

vitaly-zdanevich:


jjjake:
Thanks for the contribution, however we do not recommend installing `internetarchive` or `ia` via a package manager. [The recommended ways to install](https://repology.org/project/python:internetarchive/versions) `internetarchive` or `ia` is via pip, installing from source (via github/pypi), or using a binary.

jjjake:
@vitaly-zdanevich As you'll notice on https://repology.org/project/python:internetarchive/versions, packages created by third parties tend to drift a great deal.

https://pypi.org/project/internetarchive/ and https://github.com/jjjake/internetarchive/ is the only place we officially release packages and the only place we can guarantee is up to date. 

I'm glad to see others are creating other packages if they find it useful, but we only officially support packages released on pypi and github at the moment. And with that in mind, it's the only place we point to in the official documentation.

PR #650, opened 2024-05-26
Skip early when downloading existing file

ChlodAlejandro:
#614 moved some skip checks to until after response headers have been received, which drastically slows down the download process if the file already exists or the file has an equal checksum.

Since whether existing files (`--ignore-existing`) or matching checksums (`--checksum`) are to be skipped, file name, and checksum are all already known prior to needing information about the file through the `Last-Modified` header, these checks should remain at the start to avoid having to make a request which would eventually be discarded anyway. This speeds up `ia download` by skipping the (relatively long-running) blocking HTTP request and also stops the script from making numerous wasted requests to the Internet Archive, especially for runs which cover hundreds of files.

### DIFF
--- a/internetarchive/files.py
+++ b/internetarchive/files.py
@@ -226,6 +226,25 @@ class File(BaseFile):
 
         parent_dir = os.path.dirname(file_path)
 
+        # Check if we should skip...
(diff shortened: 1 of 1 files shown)


jjjake:
Looks good, thank you @ChlodAlejandro!

PR #680, opened 2025-02-15
Fixed typo in files.py, corrected typings for `Item.get_files` function.

karpour:
Based on the code, both `glob_pattern` and `exclude_pattern` in `Item.get_files` should both be of type `str | list[str] | None` rather than `str | None`


jjjake:
Thanks, karpour!

PR #684, opened 2025-03-01
add ability to log to stdout

wgwz:
i find myself not wanting to have to log to a file, and thought this could be a decent approach to it.
basically the idea is to add a new line into the log config that switches from a file handler to a stream handler.
i have not tested this thoroughly yet but i think it's most of the way there.
i do plan to test this some time in the next weeks.
i thought i'd at least open a PR up for some initial review/feedback.

this PRs approach is sort of "first do least harm".
but there is potentially another way to do it, which is to actually pull the logging functionality out of the library.
and leave it to the user to configure there own logging.
this might be a undesirable change but it is an approach i have experienced in other libraries.

i.e. you set up a root logger and gather up the logs from the libraries you care to have logs for:

Refs:
- https://flask.palletsprojects.com/en/stable/logging/#other-libraries
- https://docs.sqlalchemy.org/en/20/core/engines.html#configuring-logging (example of how to connect to sqlalchemy's logging)
- https://stackoverflow.com/a/16337639/4028706

for added context, it's not so nice to have this happen when you deploy something (with the logging configured):
```
PermissionError: [Errno 13] Permission denied: '/internetarchive.log'
```

curious to hear thoughts on this, thanks for your work on this project!

### DIFF
--- a/internetarchive/session.py
+++ b/internetarchive/session.py
@@ -130,12 +130,17 @@ class ArchiveSession(requests.sessions.Session):
 
         logging_config = self.config.get('logging', {})
         if logging_config.get('level'):
-            self.set_file_logger(logging_config.get('level', 'NOTSET'),
-                                 logging_config.get('file', 'internetarchive.log'))
-            if debug or (logger.level <= 10):
+            if logging_config.get('log_to_stdout'):
+                self.set_stream_logger(logging_config.get('level', 'NOTSET'))
+                if debug or (logger.level <= 10):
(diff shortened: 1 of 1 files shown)


wgwz:
Seems to be working fine with this, for example:

```python
from internetarchive import get_session

identifier = "brian-lehrer-show-stream-2025-03-07"
config = {
    "logging": {"level": "DEBUG", "log_to_stdout": True},
    "s3": {"access": "", "secret": ""},
}
s = get_session(config=config)
item = s.get_item(identifier)
print(item)
```

PR #688, opened 2025-03-19
first pass at 'ia simplelists'

jjjake:


jjjake (commented):
@day We can skip the `supbrocess` import as well as `urllib.parse` with something like:

```
if args.user:
    flag_user = args.user
else:
    flag_user = args.session.config.get("general", {}).get("screenname")
```

day (commented):
I have only tested the flag subcommand and not simplelists.

internetarchive/cli/ia_flag.py
```diff
@@ -22,49 +22,60 @@
 from __future__ import annotations
 
 import argparse
-import json
```
day: Removed unnecessary imports.

internetarchive/cli/ia_flag.py
```diff
@@ -74,35 +85,23 @@ def main(args: argparse.Namespace, parser: argparse.ArgumentParser) -> None:
         parser: Argument parser for error handling
     """
     item = args.session.get_item(args.identifier)
-    
     if args.user:
-        user = args.user
+        flag_user = args.user
     else:
-        # Extract the email from the cookie
```
day: Much simpler, yes. Thanks. ;-)

internetarchive/cli/ia_flag.py
```diff
@@ -22,49 +22,60 @@
...
-        "-d", "--delete-flag",
+        "-d",
+        "--delete-flag",
         metavar="CATEGORY",
         type=str,
-        help="Add identifier to specified parent list"
+        help="Delete a flag from the item",
     )
 
-    parser.set_defaults(func=lambda args: main(args, parser))
```
day: Honestly, this was probably not necessary, and I might push a reversion if this PR isn't pulled soon.

internetarchive/cli/ia_simplelists.py
```diff
@@ -0,0 +1,145 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import annotations
+
+import argparse
+import json
```
JustAnotherArchivist, line 25:
```suggestion
from internetarchive.utils import json
```

to make use of ultrajson when installed, like elsewhere in the code.

day: I tried to use the "Commit suggestion" feature here (new to me) and it made the commit under my main Github profile (probably will do the same for this comment). So, anyway, I didn't want it that way. And it was failing linter checks, so I reverted the change.

internetarchive/cli/ia_flag.py
jjjake: This line is necessary. It's how subcommands get executed with all of the parser configuration and setup from ia.py.

jjjake: I'm sorry, that was in reference to `parser.set_defaults(func=lambda args: main(args, parser))` ... The github collaboration workflow is very confusing to me. I think we should use PR's in the future.

jjjake (commented):
Looking good! I'll get the tests passing and merge.

day:
I tried again to put in the ultrajson change and it seems to work fine locally but linter complains about import order. Attempted to match apparent order preferences by looking at other files (e.g. ia_search.py) but it still complains. I installed the Ruff linter and Tox for VS Code so I wouldn't have to push to run checks but it's not complaining locally.

internetarchive/cli/ia_simplelists.py
```diff
@@ -26,6 +26,7 @@
 
 from internetarchive.utils import json
 
+
```
day, line 29: Okay. An extra line ;-) I'll install pylint too.

jjjake:
@day I believe the unsorted imports error was due to a missing empty line after the imports. I just pushed a fix for that. We have @JustAnotherArchivist's suggestion in there now and it's working.

I'll cleanup any remaining tests and get this merged.

In the future, let's use PR's rather than the collaborator feature. I think it will be easier to keep track of what's going on (at least for me :-)). Thanks for all the help!

PR #693, opened 2025-04-24
Correct behavior of ia_copy to avoid dropping path prefixes, fix ia_move

jjjake:
src file assertion splits once (removes item prefix) but actual copy operation splits without limit, so all path prefixes including subdirs are removed

this brings the behavior of the return value in line with the assertion and fixes the ia_move behavior where moved files in subdirs are never deleted


jjjake (commented):
Looks good to me. In the future we can consolidate the file path parsing from args.source.

PR #716, opened 2025-09-16
More complete percent encoding of download path (Windows) + add a traversal guard

### LINKED ISSUE #595: Files with colons in the name do not save to disk after downloading
While trying to download this collection for The Computer Chronicles, I found that a number of files were not saving to the disk. I tried redownloading and observed in the console that it was indeed downloading normally but once complete the resulting file was 0 bytes length and had no extension. 

I went to manually download these files in my browser and noticed a common trend among the affected files is that they had a colon in the name. This probably is a Windows specific issue as that character isn't allowed in file names and my browser when saving the file replaced it with an underscore. 

Collection: https://archive.org/details/Computer_Chronicles

OS: Windows Server 2019 Datacenter
Python: 3.7.16

pengowray:
This patch does more complete percent encoding on Windows platforms, so files can still be downloaded which previously resulted in an error, or in some cases froze the console. There's also an explicit directory traversal guard added for all platforms (which should hopefully not be triggerable)

```sh
## Example of previous output (5.5.1 or 5.6.0.dev1):

> ia download test-upload-aux.txt
test-upload-aux.txt:
 downloading AUX.TXT: 27.0iB [00:00, 24.2kiB/s]
## (at this point ia freezes, and the user can't even ctrl-c)
```

## Demo

```sh
## Example output with this patch

> ia download test-upload-bad
test-upload-bad:
windows path sanitized: test-upload-bad\AUX -> test-upload-bad\AU%58
 downloading AUX: 100%|█████████████████████████████████████████| 10.0/10.0 [00:00<00:00, 8.25kiB/s]
windows path sanitized: test-upload-bad\aaa : \ | ? * > < .txt -> test-upload-bad\aaa %3A %5C %7C %3F %2A %3E %3C .txt
 downloading aaa : \ | ? * > < .txt: 100%|██████████████████████| 6.00/6.00 [00:00<00:00, 6.72kiB/s]
...

> ls .\test-upload-bad\    
...
Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
-a---          2025-09-06    10:46              6 aaa %3A %5C %7C %3F %2A %3E %3C .txt
-a---          2025-09-06    10:46             10 AU%58
...
```

# More info

I've also added a general directory traversal check to guard against such things happening in future, though hopefully there's no way to trigger it now.

For legacy reasons, files on Windows can't named certain things: `CON`, `PRN`, `AUX`, `NUL`, `LPT[1-9]`, `COM[1-9]`. This also goes for folders and includes files with extensions, e.g. in Windows 10:

```cmd
> mkdir LPT1.TXT
The directory name is invalid.
```

This patch stops these filenames freezing ia in two ways:
1. percent encodes a letter of Windows device names: AUX -> AU%58; CON -> CO%4E; COM1.txt -> COM%31.txt
2. an explicit directory traversal check, which gives an error for that file if it fails. This should error should never be reached but it's there in case. But for example, before the above percent encoding was implemented it would catch these files instead of freezing:

```sh
## Examples of directory traversal guard getting triggered in test build without percent-encoding

> ia download test-upload-aux.txt
test-upload-aux.txt:
 error: Unsafe file path resolved outside destination directory: test-upload-aux.txt\AUX.TXT
 downloading test-upload-aux.txt_archive.torrent: 100%|██████████████████████████████████████████████| 1.66k/1.66k [00:00<00:00, 2.29MiB/s]
 downloading test-upload-aux.txt_files.xml: 1.45kiB [00:00, 1.09MiB/s]
 downloading test-upload-aux.txt_meta.sqlite: 100%|██████████████████████████████████████████████████| 20.0k/20.0k [00:00<00:00, 13.9MiB/s]
 downloading test-upload-aux.txt_meta.xml: 530iB [00:00, 603kiB/s]
  
> ia download AUX.TXT        
AUX.TXT:
 error: Unsafe file path resolved outside destination directory: AUX.TXT\AUX.TXT_archive.torrent
 error: Unsafe file path resolved outside destination directory: AUX.TXT\AUX.TXT_files.xml
 error: Unsafe file path resolved outside destination directory: AUX.TXT\AUX.TXT_meta.sqlite
 error: Unsafe file path resolved outside destination directory: AUX.TXT\AUX.TXT_meta.xml
 error downloading file AUX.TXT\device-name-collisions-test.txt, exception raised: [WinError 267] The directory name is invalid: 'AUX.TXT'

## In the previous version, both of the above would cause ia to freeze.
```

As the above example shows, this sanitation is needed on identifiers too, e.g. to allow downloading of [archive.org/details/AUX.TXT](https://archive.org/details/AUX.TXT).

Files ending in dots and/or spaces are also url encoded rather than failing.

The other thing this patch has added is percent-encoding of backslashes (`\`) in filenames. On the Internet Archive they are always part of a filename and never a path separator, so when downloaded they are percent-encoded too rather than creating paths.

## Security fixes?

- Untrusted filenames matching Windows reserved devices (COM1, LPT1, etc): Low. 
- Backslash to directory separator: Low.

In the version previous to this patch, there might be some way for a malicious uploader to exploit the behavior of COM1, LPT1, etc, but it seems unlikely. The victim would have to download the malicious item while they have an exploitable device connected to one of these typically unused ports. Perhaps the attacker could print something on an attached printer, but I have not investigated closely enough to see if this is a real threat.

Backslashes in filenames of downloaded files were still treated as a directory separators, but other safeguards were put it place to prevent directory traversal in 5.5.1.

## Alternatives:

- I believe there is an API or some way on Windows which _does_ allow the creation of these reserved names (e.g. `mkdir AUX` fails but `touch AUX` succeeds). I haven't investigated how to do this in Python or in general. When such files are created, they can be a pain. e.g. It looks like you can't delete a file named `AUX` in Explorer without admin permissions. So I've gone with percent-encoding of a letter in the filename, which seems more robust than getting around the limitation another way.

## Things not done:

- I don't think the percent _decoding_ is done when uploading
- I have tested the new app version but I have not tested the actual python tests (I got a bit lost trying to work out how to run them correctly)
- I haven't done extensive testing of the >240 character length warning
- Have not done tests with various different parameters like --no-directories, --destdir, etc

## Summary:

- Allows downloading of any filename by doing more complete percent encoding (Windows)
- Adds a directory traversal guard when downloading (all platforms)
- Adds a warning when downloading files with a path > 240 characters (Windows) 

Also this resolves #595

### DIFF
--- a/internetarchive/exceptions.py
+++ b/internetarchive/exceptions.py
@@ -54,3 +54,8 @@ class AccountAPIError(Exception):
     def __init__(self, message: str, error_data: Optional[Dict] = None):
         super().__init__(message)
         self.error_data = error_data
+
+
+class DirectoryTraversalError(Exception):
+    """Raised when a computed local file path escapes the intended destination directory."""
+    pass
--- a/internetarchive/files.py
+++ b/internetarchive/files.py
@@ -233,27 +233,24 @@ class File(BaseFile):
         self.item.session.mount_http_adapter(max_retries=retries)
         file_path = file_path or self.name
 
+        if os.name == 'nt' and not return_responses:
+            file_path, _ = utils.sanitize_windows_relpath(
+                file_path,
+                verbose=bool(verbose),
+                printer=lambda m: print(m, file=sys.stderr),
+            )
+
         if destdir:
             if return_responses is not True:
                 try:
-                    os.mkdir(destdir)
-                except FileExistsError:
+                    os.makedirs(destdir, exist_ok=True)
+                except OSError:
                     pass
             if os.path.isfile(destdir):
                 raise OSError(f'{destdir} is not a directory!')
             file_path = os.path.join(destdir, file_path)
 
-        # Windows-only local filename sanitization for invalid / reserved names.
-        # Keep original remote filename (self.name) unchanged for HTTP request.
-        # Only sanitize the final path component, not the whole provided path.
-        if os.name == 'nt' and not return_responses:
-            base_dir, base_name = os.path.split(file_path)
-            sanitized, modified = utils.sanitize_windows_filename(base_name)
-            if modified:
-                if verbose:
-                    print(f' encoding windows filename component: {base_name} -> {sanitized}', file=sys.stderr)
-                log.debug('Sanitized Windows filename component %r -> %r', base_name, sanitized)
-                file_path = os.path.join(base_dir, sanitized)
+    # Windows sanitization handled earlier; legacy comment removed.
 
         # Directory traversal guard (all platforms). Ensure target path is inside destdir (or cwd if none provided).
         # Determine intended base directory.
--- a/internetarchive/files.py
+++ b/internetarchive/files.py
@@ -243,7 +243,37 @@ class File(BaseFile):
                 raise OSError(f'{destdir} is not a directory!')
             file_path = os.path.join(destdir, file_path)
 
+        # Windows-only local filename sanitization for invalid / reserved names.
+        # Keep original remote filename (self.name) unchanged for HTTP request.
+        # Only sanitize the final path component, not the whole provided path.
+        if os.name == 'nt' and not return_responses:
+            base_dir, base_name = os.path.split(file_path)
+            sanitized, modified = utils.sanitize_windows_filename(base_name)
+            if modified:
+                if verbose:
+                    print(f' encoding windows filename component: {base_name} -> {sanitized}', file=sys.stderr)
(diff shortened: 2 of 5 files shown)


jjjake (approved):
Looks good, thanks so much @pengowray!

jjjake:
I created #719 to cleanup some tests, FYI. I'll give you time to review that if you'd like, but no worries if not. I'll merge it in a few days. Thanks again for the contributions, I appreciate it!

pengowray:
No problems. No need for me to review further changes. I appreciate you handling it

PR #742, opened 2026-02-04
Update cli.rst

hetzenmat:
I noticed that `ia help CMD` no longer works.
The correct command would be `ia CMD --help`.


jjjake (approved):
Looks good, thank you!

PR #780, opened 2026-06-29
Rename `ia tasks --follow-task-log` short flag -F -> -f (5.10.1)

jjjake:
## Summary

`ia tasks --follow-task-log` used `-F`; renamed to `-f` to match `tail -f` (the universal "follow" mnemonic). `-f` was unused in the tasks subcommand. `-F` shipped only in 5.10.0 (same day), so it's dropped outright rather than aliased.

Cuts a 5.10.1 patch release.

## Changes
- `ia_tasks.py`: `-F` -> `-f` (the option + the `-G/-f` identifier-shadow error message).
- `tests/cli/test_ia_tasks.py`, `docs/source/cli.rst`: updated to `-f`.
- `HISTORY.rst`: 5.10.1 entry; `__version__` -> `5.10.1`.

## Verification
- `ruff check`: clean
- `pytest tests/cli/test_ia_tasks.py`: 11 passed
- `ia tasks --help` shows `-f, --follow-task-log`

## Release
After merge: `git checkout master && git pull && make publish` -> 5.10.1.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


vxbinaca:
We ran into a collision with single letter flags and eventually will transition to double-tack full name tags. But cool you changed that.

jjjake:
Yes, the single letter flags can get confusing when trying to avoid conflicts and stay consistent across sub-commands!

I had a last minute change of heart on this specific one, and thought I'd change it immediately after release to avoid pain/confusion in the future (i.e. I thought this `tail -f`-like feature would make for better discovery and stick in people's memory better than the `-F` (initially chosen for consistency with `ia tasks -G`)).

Transitioning to double-tack full name tags definitely sounds like a good idea to me. Let me know if you have any other feedback on this, and sorry for any trouble my short option naming conventions may have caused you!

vxbinaca:
no no we do library calls we don't care really and with my uses I use double-tack flags, just wanted to pass along some sage advice about flag conventions because we want to on our end kill single-tack flags, because theres a naming space problem in the alphabet.

You're good though do you. And thank you, Jake. Thank you.

